PR #16, opened 2021-05-27
Improve reliability of Zayo parser and add another test file for it

glennmatthews:
- Add another lightly anonymized example Zayo notification email to integration tests
- Some minor code cleanup and a few TODOs to possibly revisit in a future PR
- Improve Zayo parser resilience:
  - Use [quopri](https://docs.python.org/3/library/quopri.html) to handle the `quoted-printable` encoding added to the original HTML content
  - Handle a few alternate cases that were initially causing exceptions to be thrown when I first attempted parsing this new file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
-## v1.0.3
+## v1.1.0
+
+### Added
+
+- #16 - changed `MaintenanceNotification.raw` from `str` to `bytes`
 
 ## v1.0.2 - 2021-05-05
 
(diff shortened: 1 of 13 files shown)


chadell (dismissed):
LGTM as it is, but I would encourage you to actually do as you suggested and change `MaintenanceNotification.raw` from `str` to `bytes` if this simplifies your PR, or if not, to create an issue to do it sooner than later, now that the library usage is still limited.

chadell (approved):
LGTM, we will need to careful use it from the Plugin to match the new raw type.

chadell:
@glennmatthews would you like to add this PR in the CHANGELOG.md, pointing to release 1.1.0 (because the change with `raw`)

chadell (approved):
LGTM

PR #17, opened 2021-05-31
Add Lumen first parser test and refactor HTML class

chadell:
- [x] Add test cases for Lumen maintenances (we still need more samples to check multiples status, etc.)
- [x] Refactor HTML `process` adding a new `parse_html` function that is the one to be implemented by the child classes, so decreasing repeated code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,12 @@ You can leverage on this library in your automation framework to process circuit
 ## Supported Providers
 
 - EuNetworks
+- Lumen\*
 - NTT
 - PacketFabric
-- Zayo
+- Zayo\*
+
+> Note\*: These providers are not supporting the BCOP standard natively, thus maybe they have some gaps on the implemented parser that will be refined with new test cases.
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -23,13 +23,18 @@ You can leverage on this library in your automation framework to process circuit
 
 ## Supported Providers
 
+### Supported providers using the BCOP standard
+
 - EuNetworks
-- Lumen\*
 - NTT
 - PacketFabric
-- Zayo\*
 
-> Note\*: These providers are not supporting the BCOP standard natively, thus maybe they have some gaps on the implemented parser that will be refined with new test cases.
+### Supported providers based on other parsers
+
+- Lumen
+- Zayo
+
+> Note: Because these providers do not support the BCOP standard natively, maybe there are some gaps on the implemented parser that will be refined with new test cases. We encourage you to report related **issues**!
 
 ## Installation
 
--- a/circuit_maintenance_parser/parser.py
+++ b/circuit_maintenance_parser/parser.py
@@ -4,7 +4,7 @@ import base64
 import calendar
 import datetime
 import quopri
-from typing import Iterable, Union, Dict
+from typing import Iterable, Union, Dict, Mapping
 
 import bs4  # type: ignore
 from bs4.element import ResultSet  # type: ignore
@@ -180,21 +180,26 @@ class Html(MaintenanceNotification):
 
     def process(self) -> Iterable[Maintenance]:
         """Execute parsing."""
-        data: Dict[str, Union[int, str, Iterable]] = {
+        result = []
+
+        data_base: Dict[str, Union[int, str, Iterable]] = {
             "provider": self._default_provider,
             "organizer": self._default_organizer,
         }
         try:
             soup = bs4.BeautifulSoup(quopri.decodestring(self.raw), features="lxml")
 
-            self.parse_html(soup, data)
+            # Even we have not noticed any HTML notification with more than one maintenance yes, we define the
+            # return of `parse_html` as a multiple object, even for now we only return a list of one object.
+            for data in self.parse_html(soup, data_base):
+                result.append(Maintenance(**data))
 
-            return [Maintenance(**data)]
+            return result
 
         except ValidationError as exc:
             raise MissingMandatoryFields from exc
 
-    def parse_html(self, soup: ResultSet, data: Dict):
+    def parse_html(self, soup: ResultSet, data_base: Dict) -> Iterable[Union[Mapping[str, Union[str, int, Dict]]]]:
         """Custom HTML parsing."""
         raise NotImplementedError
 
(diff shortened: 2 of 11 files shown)


chadell:
@glennmatthews I plan to add more use-cases to Lumen to make it more complete, but because of the small refactor on the `HTML` class raised this PR for a early review.

README.md
```diff
@@ -24,9 +24,12 @@ You can leverage on this library in your automation framework to process circuit
 ## Supported Providers
 
 - EuNetworks
+- Lumen\*
 - NTT
 - PacketFabric
-- Zayo
+- Zayo\*
+
+> Note\*: These providers are not supporting the BCOP standard natively, thus maybe they have some gaps on the implemented parser that will be refined with new test cases.
```
glennmatthews, line 32: Wonder if it would make sense to split this list into two lists, i.e. "Supported providers using the BCOP standard" and "Supported providers based on other parsers"?

circuit_maintenance_parser/parser.py
```diff
@@ -175,7 +179,23 @@ class Html(MaintenanceNotification):
...
+        data: Dict[str, Union[int, str, Iterable]] = {
+            "provider": self._default_provider,
+            "organizer": self._default_organizer,
+        }
+        try:
+            soup = bs4.BeautifulSoup(quopri.decodestring(self.raw), features="lxml")
+
+            self.parse_html(soup, data)
+
+            return [Maintenance(**data)]
```
glennmatthews, line 192: I like moving the `quopri.decodestring()` to the base class, but I wonder if constructing the `[Maintenance(**data)]` in the base class is overly limiting - do we anticipate any case where a single `self.raw` would instead result in more than one `Maintenance` object being constructed?

circuit_maintenance_parser/parsers/lumen.py
```diff
@@ -0,0 +1,100 @@
...
+            self.parse_tables(soup.find_all("table"), data)
+
+        except Exception as exc:
+            raise ParsingError from exc
+
+    def parse_spans(self, spans: ResultSet, data: Dict):
+        """Parse Span tag."""
+        for line in spans:
+            if isinstance(line, bs4.element.Tag):
+                if line.text.lower().strip().startswith("scheduled maintenance window #:"):
```
glennmatthews, line 40: Consider creating an intermediate variable like `text = line.text.lower().strip()` so that you don't have to keep calling this chain of functions over and over again.

circuit_maintenance_parser/parsers/lumen.py
```diff
@@ -0,0 +1,100 @@
...
+                    data["maintenance_id"] = line.text.lower().strip().split("#: ")[-1]
+                elif line.text.lower().strip().startswith("summary:"):
+                    for sibling in line.next_siblings:
+                        text = sibling.text if isinstance(sibling, bs4.element.Tag) else sibling
```
glennmatthews, line 44: Similarly, save yourself from needing to call `strip()` over and over again:

```suggestion
                        text = sibling.text.strip() if isinstance(sibling, bs4.element.Tag) else sibling.strip()
```

circuit_maintenance_parser/parsers/lumen.py
```diff
@@ -0,0 +1,100 @@
...
+                        if text.strip() != "":
+                            data["summary"] = text.strip()
+                            break
+                elif line.text.lower().strip().startswith("updates:"):
+                    for sibling in line.next_siblings:
+                        text = sibling.text if isinstance(sibling, bs4.element.Tag) else sibling
```
glennmatthews, line 50:
```suggestion
                        text = sibling.text.strip() if isinstance(sibling, bs4.element.Tag) else sibling.strip()
```

circuit_maintenance_parser/parsers/lumen.py
```diff
@@ -0,0 +1,100 @@
...
+    def parse_tables(self, tables: ResultSet, data: Dict):
+        """Parse Table tag."""
+        circuits = []
+        for table in tables:
+            cells = table.find_all("td")
+            if cells[0].string == "Start" and cells[1].string == "End":
+                num_columns = 2
+                num_rows = int(len(cells) / num_columns - 1)
+                for idx in range(1, num_rows + 1):
+                    if "GMT" in cells[idx * num_columns].string and "GMT" in cells[idx * num_columns + 1].string:
```
glennmatthews, line 68: Perhaps iterate by `num_columns` as a step so you don't need to keep repeatedly doing `idx * num_columns`?

```suggestion
                for idx in range(num_columns, len(cells), num_columns):
                    if "GMT" in cells[idx].string and "GMT" in cells[idx + 1].string:
```
... etc.

circuit_maintenance_parser/parsers/lumen.py
```diff
@@ -0,0 +1,100 @@
...
+                num_columns = 1
+                if len(cells) % 10 == 0:
+                    num_columns = 10
+                elif len(cells) % 9 == 0:
+                    num_columns = 9
+                else:
+                    logger.error("Unexpected table format: %s", cells)
+
+                num_rows = int(len(cells) / num_columns - 1)
+                for idx in range(1, num_rows + 1):
```
glennmatthews, line 86: Similarly:

```suggestion
                for idx in range(num_columns, len(cells), num_columns):
```

glennmatthews (approved):

circuit_maintenance_parser/parser.py
chadell: I wanted to not limit a notification to only one maintenance even all the notifications we have seen so far contained only one maintenance, but being able to get ready for the more generic use case that contains more than one maintenance.
But maybe you are right and we could simplify it to one `raw` to one `Maintenance`, and now that we are in early stage of the project is a good time to do it...

glennmatthews: I actually meant things the other way round - if we do want to permit the case where more than one maintenance are returned, the code you have here as written is overly restrictive; I would want to let `parse_html` return a list of (one or more) `Maintenance` objects itself rather than hard-coding things to always construct exactly one such object.

chadell: gotcha.
I started first with iCalendar, that supports multiple EVENTS in one notifications, so I hardcoded the same signature behaviour for HTML even I have never seen any one with this behaviour, so I ended up with always returning a list of 1 element, as you noticed.
To make this code ready to accommodate this use-case, I have changed a bit the method to handle future HTML parsers that could contain multiple maintenances.

PR #18, opened 2021-06-03
Add Telstra parser and make 'stamp' attribute optional

chadell:
* Add Telstra HTML parser
* Make `stamp` attribute optional, as it's not available in some HTML parsers


circuit_maintenance_parser/output.py
```diff
@@ -132,9 +132,10 @@ class Maintenance(BaseModel):
     status: Status
     start: StrictInt
     end: StrictInt
-    stamp: StrictInt
     organizer: StrictStr
 
+    # Non mandatory attributes
+    stamp: Union[StrictInt, None] = None
```
glennmatthews, line 138: Simpler option:

```suggestion
    stamp: Optional[StrictInt] = None
```

circuit_maintenance_parser/parsers/telstra.py
```diff
@@ -0,0 +1,78 @@
...
+                    data["status"] = Status("CONFIRMED")
+                else:
+                    continue
+                break
+            for th_element in table.find_all("th"):
+                if not th_element.string:
+                    continue
+                th_text = th_element.string.strip()
+                if th_text == "To:":
+                    data["account"] = th_element.next_sibling.next_sibling.string
```
glennmatthews, line 54: Might be nice to assign `th_element.next_sibling.next_sibling.string` to a variable.
```suggestion
                th_text = th_element.string.strip()
                td_text = th_element.next_sibling.next_sibling.string
                if th_text == "To:":
                    data["account"] = td_text
```

(etc.)

Additionally, would something like `th_element.next_element.string` be correct as well?

circuit_maintenance_parser/parsers/telstra.py
```diff
@@ -0,0 +1,78 @@
...
+                        data["circuits"].append(CircuitImpact(impact=Impact("OUTAGE"), circuit_id=circuit_id))
+                elif th_text == "Maintenance Details:":
+                    sentences: List[str] = []
+                    for element in th_element.next_elements:
+                        if element.string == "Service(s) Impacted:":
+                            break
+                        if element.string and element.string not in ["\n", "", "\xa0"] + sentences:
+                            sentences.append(element.string)
+                    if sentences:
+                        data["summary"] = ". ".join(sentences[1:])
```
glennmatthews, line 77: Why do we discard the first sentence? Mind adding a comment?

circuit_maintenance_parser/parsers/telstra.py
chadell: Not sure what do you exactly mean with "Additionally, would something like th_element.next_element.string be correct as well?" 
but if you mean " td_text = th_element.next_element.string" this doesn't work because the HTML structure, and depending on the level it is the info I use next_sibling (same level) or next_element (whatever comes afterwards)

glennmatthews: Yeah, that's what I meant, sorry for the lack of clarity. Thanks for the explanation.

tests/integration/test_parsers.py
```diff
@@ -0,0 +1,113 @@
...
+        (ICal, Path(dir_path, "data", "ical", "ical_no_stamp"), ParsingError,),
+        (ICal, Path(dir_path, "data", "ical", "ical_no_start"), ParsingError,),
+        (ICal, Path(dir_path, "data", "ical", "ical_no_end"), ParsingError,),
+        # Zayo
+        (ParserZayo, Path(dir_path, "data", "zayo", "zayo_missing_maintenance_id.html"), MissingMandatoryFields,),
+        (ParserZayo, Path(dir_path, "data", "zayo", "zayo_bad_html.html"), MissingMandatoryFields,),
+    ],
+)
+def test_errored_parsing(parser, raw_file, exception):
+    """Tests for Ical parser."""
```
glennmatthews, line 108:
```suggestion
    """Negative tests for various parsers."""
```

tests/integration/test_parsers.py
```diff
@@ -0,0 +1,113 @@
...
+            ParserZayo,
+            Path(dir_path, "data", "zayo", "zayo2.html"),
+            Path(dir_path, "data", "zayo", "zayo2_result.json"),
+        ),
+    ],
+)
+def test_complete_parsing(parser, raw_file, results_file):
+    """Tests parser."""
+    with open(raw_file, "rb") as file_obj:
+        parser = parser(raw=file_obj.read())
```
glennmatthews, line 83: I don't like redefining the `parser` variable like this. Maybe rename the function parameter to `parser_class` and reserve `parser` for the instance you're creating here? Same comment on the other test function below.

glennmatthews (dismissed):

tests/integration/test_parsers.py
chadell: absolutely, I did a quick replace and I missed that point

PR #19, opened 2021-06-04
Add Megaport parser

chadell:


circuit_maintenance_parser/output.py
```diff
@@ -132,9 +132,10 @@ class Maintenance(BaseModel):
     status: Status
     start: StrictInt
     end: StrictInt
-    stamp: StrictInt
     organizer: StrictStr
 
+    # Non mandatory attributes
+    stamp: Union[StrictInt, None] = None
```
glennmatthews, line 138: As in #18, suggest using `Optional` here rather than `Union[..., None]`

circuit_maintenance_parser/parsers/megaport.py
```diff
@@ -0,0 +1,68 @@
...
+
+            circuit_table = tr_elem.find("table")
+            if circuit_table and circuit_table.find("th").string == "Service ID":
+                data["circuits"] = []
+                num_columns = len(circuit_table.find_all("th"))
+                cells = circuit_table.find_all("td")
+                for idx in range(0, len(cells), num_columns):
+                    data["circuits"].append(CircuitImpact(impact=Impact("OUTAGE"), circuit_id=cells[idx].a.string))
+
+                return
```
glennmatthews, line 68: Unnecessary `return` statement here.

circuit_maintenance_parser/parsers/megaport.py
```diff
@@ -0,0 +1,68 @@
...
+                p_text = p_elem.text
+                if not p_text:
+                    continue
+                if p_text.startswith("This is a reminder that Planned maintenance"):
+                    data["maintenance_id"] = p_elem.find("b").string
+                    data["status"] = Status("CONFIRMED")
+                elif p_text.startswith("Hi "):
+                    data["account"] = p_text.split("Hi ")[-1]
+                elif p_text.startswith("Purpose of Maintenance:"):
+                    data["summary"] = p_text.split("Purpose of Maintenance: ")[-1]
```
glennmatthews, line 52: Just a general thought, but I think in the long term we might be better to use regex matches instead of `str.split()` to extract interesting substrings. Not blocking; we can always refactor this later.

glennmatthews (dismissed):

circuit_maintenance_parser/output.py
chadell: sure, It's fixed after the merge from that PR (I only cherry-picked my original commit)

circuit_maintenance_parser/parsers/megaport.py
chadell: I will give it a try

circuit_maintenance_parser/parsers/megaport.py
chadell: 😄 , not actually, without the return, the `for` loop continues and there are objects that can't do a `find_all`... so I should break the loop once I'm "done". 
If we could get a nicer approach I'm all ears

glennmatthews: Ah, I missed that nuance. In that case, perhaps `break` and a comment would be clearer?

PR #20, opened 2021-06-07
Release v1.1.0

chadell:


.travis.yml
```diff
@@ -55,8 +55,6 @@ jobs:
         skip_cleanup: true
         "on":
           tags: true
-          branch: "main"
-          python: 3.7
```
glennmatthews, line 59: Is the python version check still needed? IIRC without it these steps (`deploy` and `deploy-pypi`) will run for each python version in the matrix, instead of once only for python 3.7.

glennmatthews (approved):

.travis.yml
dgarros: as far as I remember only the stage `test` will run for all versions of python defined
We have something similar for the schema-enforcer https://github.com/networktocode/schema-enforcer/blob/develop/.travis.yml

glennmatthews: You're right. I just checked the schema-enforcer last Travis run on `main` and it did only execute `test` against the matrix. Thanks for the information!

PR #25, opened 2021-06-10
Cogent parser and integration tests

pke11y:
- Cogent parser based on `Planned Network Maintenance` notifications only
- Integration tests included for :

1. Planned Network Maintenance
2. Planned Network Maintenance Correction


pke11y:
I've included `cogent3.html` as a reference. It did not work with the parser. The `bs4` library could only parse the `html head` section. The `html body` was not available for processing.

chadell (changes requested):
Hi @pke11y , great job!
Extra points:
* In order to solve your issue with HTML3 change this [line](https://github.com/networktocode/circuit-maintenance-parser/blob/develop/circuit_maintenance_parser/parser.py#L190) to `bs4.BeautifulSoup(quopri.decodestring(self.raw), 'html.parser')`
* Update the Readme with the new parser
* Update the [__init__.py ](https://github.com/networktocode/circuit-maintenance-parser/blob/develop/circuit_maintenance_parser/__init__.py)
* Update Changelog

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,75 @@
...
+                        data["status"] = Status("CONFIRMED")
+                    else:
+                        data["status"] = Status("TENTATIVE")
+                    data["summary"] = line
+                elif line.startswith("Dear"):
+                    match = re.search("Dear (.*),", line)
+                    if match:
+                        data["account"] = match.group(1)
+                    else:
+                        data["account"] = "Cogent Customer"
```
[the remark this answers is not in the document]
pke11y: Ok, so just what we match against and remove the `else`

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,75 @@
...
+
+        except Exception as exc:
+            raise ParsingError from exc
+
+    def parse_br(self, br_results: ResultSet, data: Dict):
+        """Parse <br> tag."""
+        for br in br_results:
+            div_text = br.parent.text
+            for line in re.split(r"\n", div_text):
+                if line.endswith("Network Maintenance"):
```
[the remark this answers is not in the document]
pke11y: Will add an additional check for `Correction` and set status to `RE-SCHEDULE`

circuit_maintenance_parser/parsers/cogent.py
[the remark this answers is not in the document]
pke11y: Might be useful to be able to specify a `parser` to use in `bs4`

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,80 @@
...
+                elif line.startswith("Dear"):
+                    match = re.search("Dear (.*),", line)
+                    if match:
+                        data["account"] = match.group(1)
+                elif line.startswith("Start time:"):
+                    match = re.search("Start time: (.*) \(local time\) (\d+/\d+/\d+)", line)
+                    if match:
+                        start_str = " ".join(match.groups())
+                        start = datetime.strptime(start_str, "%I:%M %p %d/%m/%Y")
+                    data["start"] = self.dt2ts(start)
```
chadell, line 54:
```suggestion
                        data["start"] = self.dt2ts(start)
```

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,80 @@
...
+                elif line.startswith("End time:"):
+                    match = re.search("End time: (.*) \(local time\) (\d+/\d+/\d+)", line)
+                    if match:
+                        end_str = " ".join(match.groups())
+                        end = datetime.strptime(end_str, "%I:%M %p %d/%m/%Y")
+                    data["end"] = self.dt2ts(end)
```
chadell, line 60:
```suggestion
                        data["end"] = self.dt2ts(end)
```

README.md
```diff
@@ -35,6 +35,7 @@ You can leverage on this library in your automation framework to process circuit
 - Megaport
 - Telstra
 - Zayo
+- Cogent
```
glennmatthews, line 38: I guess let's try to keep these lists alphabetized?

circuit_maintenance_parser/__init__.py
```diff
@@ -22,6 +23,7 @@
     ParserTelstra,
     ParserZayo,
     ICal,
+    ParserCogent,
```
glennmatthews, line 26: Let's alphabetize this list as well, I think.

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,80 @@
...
+            return [data]
+
+        except Exception as exc:
+            raise ParsingError from exc
+
+    def parse_br(self, br_results: ResultSet, data: Dict):
+        """Parse <br> tag."""
+        for br in br_results:
+            div_text = br.parent.text
+            for line in re.split(r"\n", div_text):
```
glennmatthews, line 42: Would `for line in div_text.splitlines()` work just as well?

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,80 @@
...
+                elif line.startswith("Dear"):
+                    match = re.search("Dear (.*),", line)
+                    if match:
+                        data["account"] = match.group(1)
+                elif line.startswith("Start time:"):
+                    match = re.search("Start time: (.*) \(local time\) (\d+/\d+/\d+)", line)
+                    if match:
+                        start_str = " ".join(match.groups())
+                        start = datetime.strptime(start_str, "%I:%M %p %d/%m/%Y")
+                        data["start"] = self.dt2ts(start)
```
glennmatthews, line 54: I see that the string being matched here references `(local time)` rather than UTC. Do we need to account for time zones in this parser in some way?

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,80 @@
...
+                elif line.startswith("Work order number:"):
+                    match = re.search("Work order number: (.*)\s+", line)
+                    if match:
+                        data["maintenance_id"] = match.group(1)
+                elif line.startswith("Order ID(s) impacted:"):
+                    data["circuits"] = []
+                    match = re.search("Order ID\(s\) impacted: (.*)\s+", line)
+                    for circuit_id in match.group(1).split(","):
+                        data["circuits"].append(
+                            CircuitImpact(impact=Impact("OUTAGE"), circuit_id=circuit_id.lstrip().strip())
```
glennmatthews, line 70: `lstrip().strip()` looks suspect to me. Isn't the `lstrip()` redundant?

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,80 @@
...
+            self.parse_title(soup.find_all("title"), data)
+            return [data]
+
+        except Exception as exc:
+            raise ParsingError from exc
+
+    def parse_br(self, br_results: ResultSet, data: Dict):
+        """Parse <br> tag."""
+        for br in br_results:
+            div_text = br.parent.text
```
glennmatthews, line 41: Feels weird to me that we're searching for all `<br>` just so we can find the very first `<br>` and then use it to find the `<div>` that is its parent and iterate over that. Are there any distinguishing characteristics (maybe `class="a3s aiL "`?) that we can use to find the div directly?

circuit_maintenance_parser/parsers/cogent.py
chadell: yes, it is (I proposed it but not needed)

circuit_maintenance_parser/parsers/cogent.py
chadell: it should be UTC

pke11y: Ok, are you saying I need to determine what the local time is and convert to UTC?

chadell: sadly, I have to say yes :(
It's weird they to not offer a GMT/UTC time... but seeing this notification, I only see taking the location and using `pytz` convert between timezone 😞

pke11y: Added a solution for this. It's quite slow during testing. A few seconds extra, but it's noticeable. I can look for an alternative if that's going to be an issue.

pke11y: I'll need to add the additional libraries to `poetry` too

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -28,50 +28,64 @@ def parse_html(self, soup, data_base):
...
+                    match = re.search(r"[^Cogent].*?((\b[A-Z][a-z\s-]+)+, ([A-Za-z-]+[\s-]))", line)
+                    if match:
+                        parsed_timezone = city_timezone(match.group(1).strip())
+                        local_timezone = timezone(parsed_timezone)
+                        # set start time using the local city timezone
+                        start = datetime.strptime(start_str, "%I:%M %p %d/%m/%Y")
+                        local_time = local_timezone.localize(start)
+                        # set start time to UTC
+                        utc_start = local_time.astimezone(UTC)
+                        data["start"] = self.dt2ts(utc_start)
```
glennmatthews, line 69: Nice! For troubleshooting purposes, it would probably be good to include a log message or two here and below, maybe something along the lines of:

```python
logger.info("Mapped %s at %s (%s), to %s (UTC)", start_str, match.group(1), local_timezone, utc_start)
```

pke11y: Good idea Glenn. Start and End logging added:
`INFO     circuit_maintenance_parser.parsers.cogent:cogent.py:70 Mapped start time 12:00 AM 06/11/2021 at San Jose, CA (America/Los_Angeles), to 2021-11-06 07:00:00+00:00 (UTC)`
`INFO     circuit_maintenance_parser.parsers.cogent:cogent.py:83 Mapped end time 03:00 AM 06/11/2021 at San Jose, CA (America/Los_Angeles), to 2021-11-06 10:00:00+00:00 (UTC)`

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,75 @@
...
+                    data["end"] = self.dt2ts(end)
+                elif line.startswith("Work order number:"):
+                    match = re.search("Work order number: (.*)\s+", line)
+                    if match:
+                        data["maintenance_id"] = match.group(1)
+                elif line.startswith("Order ID(s) impacted:"):
+                    data["circuits"] = []
+                    match = re.search("Order ID\(s\) impacted: (.*)\s+", line)
+                    if match:
+                        data["circuits"].append(CircuitImpact(impact=Impact("OUTAGE"), circuit_id=match.group(1)))
```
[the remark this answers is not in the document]
pke11y: Added check for match as mypy was complaining.

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,110 @@
...
+                    data["circuits"] = []
+                    match = re.search(r"Order ID\(s\) impacted: (.*)", line)
+                    if match:
+                        for circuit_id in match.group(1).split(","):
+                            data["circuits"].append(
+                                CircuitImpact(impact=Impact("OUTAGE"), circuit_id=circuit_id.lstrip())
+                            )
+            break
+
+    def parse_title(self, title_results: ResultSet, data: Dict):  # pylint: disable=no-self-use
```
chadell, line 104: if `disable=no-self-use`, why not `@staticmethod`?

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,110 @@
...
+                    match = re.search(r"End time: (.*) \([A-Za-z\s]+\) (\d+/\d+/\d+)", line)
+                    if match:
+                        end_str = " ".join(match.groups())
+                elif line.startswith("Cogent customers receiving service"):
+                    match = re.search(r"[^Cogent].*?((\b[A-Z][a-z\s-]+)+, ([A-Za-z-]+[\s-]))", line)
+                    if match:
+                        parsed_timezone = city_timezone(match.group(1).strip())
+                        local_timezone = timezone(parsed_timezone)
+                        # set start time using the local city timezone
+                        start = datetime.strptime(start_str, "%I:%M %p %d/%m/%Y")
```
chadell, line 65: maybe I'm missing something, but what if the line processing for `start_str` happens later and here we still have `start_str=""`? should not  be this done in different loops?

tests/integration/test_parsers.py
```diff
@@ -87,6 +87,16 @@
...
             Path(dir_path, "data", "zayo", "zayo2_result.json"),
         ),
+        (
+            ParserCogent,
+            Path(dir_path, "data", "cogent", "cogent1.html"),
+            Path(dir_path, "data", "cogent", "cogent1_result.json"),
+        ),
+        (
+            ParserCogent,
+            Path(dir_path, "data", "cogent", "cogent2.html"),
```
chadell, line 97: I see a `cogent3.html` file but not used, any reason? if yes, maybe adding note?

circuit_maintenance_parser/utils.py
```diff
@@ -0,0 +1,16 @@
+"""Utility functions for the Circuit Maintenance parsers."""
+
+from geopy.geocoders import Nominatim  # type: ignore
+from tzwhere import tzwhere  # type: ignore
+
+
+def city_timezone(city: str) -> str:
```
chadell, line 7: I would add simple Unit test for this function

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -0,0 +1,110 @@
...
+                    match = re.search("Work order number: (.*)", line)
+                    if match:
+                        data["maintenance_id"] = match.group(1)
+                elif line.startswith("Order ID(s) impacted:"):
+                    data["circuits"] = []
+                    match = re.search(r"Order ID\(s\) impacted: (.*)", line)
+                    if match:
+                        for circuit_id in match.group(1).split(","):
+                            data["circuits"].append(
+                                CircuitImpact(impact=Impact("OUTAGE"), circuit_id=circuit_id.lstrip())
```
chadell, line 100:
```suggestion
                                CircuitImpact(impact=Impact("OUTAGE"), circuit_id=circuit_id.strip())
```
I believe this would cover your more parsing corner cases

circuit_maintenance_parser/parsers/cogent.py
chadell: maybe you could wrap up everything in a more complete function such as:
```py
def convert_time_to_utc(timezone, time):
    parsed_timezone = city_timezone(match.group(1).strip())
    local_timezone = timezone(parsed_timezone)
    local_time = local_timezone.localize(time)
    return local_time.astimezone(UTC)
# Maybe also add the logging here 
```

chadell (commented):
My only blocking concern is about the start_str processing from the timezone in cases we have not already updated the start_str in the loop (but maybe I'm missing something)

circuit_maintenance_parser/parsers/cogent.py
chadell: > maybe I'm missing something, but what if the line processing for `start_str` happens later and here we still have `start_str=""`? should not be this done in different loops?

but I see that the email format is as it is, so it happens later

chadell (approved):
LGTM, just minor comments

glennmatthews (dismissed):

circuit_maintenance_parser/parsers/cogent.py
pke11y: added @staticmethod

tests/integration/test_parsers.py
pke11y: removed

circuit_maintenance_parser/parsers/cogent.py
pke11y: updated

chadell (approved):
🚀 LGTM

Issue #24, opened 2021-06-08
Support parser versioning


chadell:
### Proposed Functionality

Parsers could support different parsing implementations for vendors.

### Use Case

When a provider offers different templates or it is transitioning, the library should provide the best parsed output.

For instance, Telstra has started using iCalendar format: https://twitter.com/NetworkAndrew/status/1402964730285789190

So, a provider parser should be able to parse multiple formats, and for each format, multiple implementations.

chadell:
Fixed in #26

PR #26, opened 2021-06-12
Refactor library to allow a Provider to concatenate multiple parsers

chadell:
This PR addresses issue #24 , so enable a Provider Parser to handle multiple parsers (in a specific order), to allow multiple versions while a provider is transitioning. This is done honouring the current signature of the library (`init_parser`).

The main change is the new class `GenericProvider` that contains all the info to parse notifications from a provider, and can concatenate multiple parsers that will be tried in order to get a valid `Maintenace`, so for example, a provider now could have in `_parser_classes` a list of `ICal`, `CustomHtml1`, `CustomHtml2`... so multiple types and versions are supported.

Other small changes:
* Added basic logging
* Removed `subject`, `sender` and `source` attributes because where unused by the library, they are more related to the client library.
* `ICal` format is shared across providers, so it's useless to run the very same test. All the providers that use it, are reusing the `ICal` parser.
* Moved tests from `integration` to `unit` because the tests are really unit tests. The Parsers alone don't know the `organizer` and `provider` so if the data is not in the notification, the result is `unknown`.  When the parser is called from the Provider, this info flows from there.


tests/unit/test_providers.py
```diff
@@ -0,0 +1,85 @@
...
+    "provider_class, raw_file, results_file",
+    [
+        # GenericProvider
+        (
+            GenericProvider,
+            Path(dir_path, "data", "ical", "ical1"),
+            Path(dir_path, "data", "ical", "ical1_result.json"),
+        ),
+        # EUNetworks
+        (EUNetworks, Path(dir_path, "data", "ntt", "ntt1"), Path(dir_path, "data", "ntt", "ntt1_result.json"),),
```
glennmatthews, line 33: Why test `ntt` data with `EUNetworks` provider? Same question regarding PacketFabric and Telstra below.

glennmatthews (commented):
Sorry for the delay in reviewing this!

tests/unit/test_providers.py
chadell: it's an error, it should use the generic ICal parser

glennmatthews (dismissed):

chadell:
thanks @glennmatthews , I will wait to merge until #25  is merged, I prefer to solve conflicts.

PR #31, opened 2021-06-22
Fix default provider when not present in notifications

chadell:
The parsed response from a Provider was not taking the proper `provider` field when it was not present in the notifications, and the default name of the class was required. 
This PR adds the proper test to catch it and the fix.

### DIFF
--- a/circuit_maintenance_parser/output.py
+++ b/circuit_maintenance_parser/output.py
@@ -8,7 +8,7 @@ from enum import Enum
 
 from typing import List, Optional
 
-from pydantic import BaseModel, validator, StrictStr, StrictInt
+from pydantic import BaseModel, validator, StrictStr, StrictInt, Extra
 
 
 class Impact(str, Enum):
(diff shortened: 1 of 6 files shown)


circuit_maintenance_parser/providers.py
```diff
@@ -68,7 +68,9 @@ def process(self, data_type: str = "") -> Iterable[Maintenance]:
         for parser_class in self._parser_classes:
             try:
                 parser = parser_class(
-                    raw=self.raw, provider_type=self.get_provider_type(), default_organizer=self._default_organizer,
+                    raw=self.raw,
+                    default_provider=self.get_provider_type(),
+                    default_organizer=self.get_default_organizer(),
```
glennmatthews, line 73: Do I understand correctly that `provider_type` is not actually an attribute on the parser class, and so it was being ignored altogether? If so, we might consider configuring `extra = 'forbid'` on our Pydantic classes (https://pydantic-docs.helpmanual.io/usage/model_config/) so that mis-specified attributes cause an error rather than being silently ignored, so as to more readily catch issues like this in the future.

chadell: that would be great!

circuit_maintenance_parser/output.py
```diff
@@ -49,7 +49,7 @@ class Status(str, Enum):
     RE_SCHEDULED = "RE-SCHEDULED"
 
 
-class CircuitImpact(BaseModel):
+class CircuitImpact(BaseModel, extra=Extra.forbid):
```
glennmatthews, line 52: Wow, I didn't know about this syntax option for specifying model configuration - that's pretty nifty!

glennmatthews (approved):

PR #32, opened 2021-06-22
Restore backwards compatibility 1.1.x

chadell:
It's built on top of #31 , only new commit is 2d224ed and adds:
* `MaintenanceNotification` class to use as adapter between old code and new features.
* `init_parser` method refactored to use the adapter class.
* `get_provider_data_type` narrowed function to only return one data type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fix
 
+- #32 - Fix backwards compatibility with version 1.1.0 that was broken in 1.2.0
 - #31 - Fix consistent provider (and other attributes) usage in all the classes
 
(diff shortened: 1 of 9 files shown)


glennmatthews (commented):
Have you tested this with the latest available release of nautobot-circuit-maintenance plugin to ensure that the backward-compatibility changes work as intended?

chadell:
> Have you tested this with the latest available release of nautobot-circuit-maintenance plugin to ensure that the backward-compatibility changes work as intended?

Yes, I did `nautobot_1  |   Nautobot initialized!`

glennmatthews (approved):

PR #41, opened 2021-08-05
Telia Parser

carbonarok:
Enable Telia parser for circuit maintenance plugin. Using ICal parsing Telia emails are very easy thanks to Christian.


circuit_maintenance_parser/providers.py
```diff
@@ -125,6 +125,13 @@ class PacketFabric(GenericProvider):
     _default_organizer = "support@packetfabric.com"
 
 
+class Telia(GenericProvider):
+    """Telia provider custom class."""
+
+    _parser_classes: Iterable[Type[Parser]] = [ICal]
+    _default_organizer = "ncm@teliacompany.com"
```
glennmatthews, line 132: Is there a reason the default organizer here doesn't match the ORGANIZER in the unit test?

glennmatthews (commented):
Nice! I hadn't realized Telia was using the standard - that's great news!

Probably should add this to the list of supported providers in the README as well.

circuit_maintenance_parser/providers.py
chadell: I believe it's because if the notification contains a organizer, the default is not used

chadell:
@carbonarok I believe that this is a great opportunity to add the dual maintenance in a notification test, would you do it?

tests/unit/test_providers.py
```diff
@@ -94,14 +109,14 @@ def test_complete_provider_process(provider_class, raw_file, results_file):
     # If the Provider test is using the GENERIC_ICAL_DATA_PATH it comes with a well-defined 'organizer'
     # from the notificaction.
     if provider_class == GenericProvider or raw_file == GENERIC_ICAL_DATA_PATH:
-        expected_result["organizer"] = "mailto:noone@example.com"
+        expected_result[0]["organizer"] = "mailto:noone@example.com"
```
chadell, line 112: @carbonarok why not just put this check into a `for loop`?

carbonarok: ah yeah good idea

chadell (dismissed):
LGTM, only point would be to add the new feature to Changelog

chadell (approved):
LGTM 🏅

glennmatthews (approved):

Issue #40, opened 2021-07-23
Difficult to troubleshoot parsing errors


glennmatthews:
### Environment

- Python version: 
- circuit_maintenance_parser version: 

### Expected Behavior

Parsing failure should result in a clearly formatted exception and/or logging message to make it easy to troubleshoot and fix.

### Observed Behavior

Due to the logic in https://github.com/networktocode/circuit-maintenance-parser/blob/develop/circuit_maintenance_parser/providers.py#L60 that catches all `ParsingError` exceptions raised by individual parser classes and discards them, the only log messages seen are generic, i.e.:

```
Parser HtmlParserCogent1 for provider Cogent was not successful
```

and the ultimately raised `ParsingError` is similarly generic:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/nautobot_circuit_maintenance/handle_notifications/handler.py", line 193, in process_raw_notification
    parsed_notifications = parser.process()
  File "/usr/local/lib/python3.7/site-packages/circuit_maintenance_parser/providers.py", line 81, in process
    raise ParsingError("None of the parsers was able to parse the notification")
circuit_maintenance_parser.errors.ParsingError: None of the parsers was able to parse the notification
```

This makes it very difficult to diagnose why the parsing failed.

### Steps to Reproduce

1.
2.
3.

chadell:
Improved in #42

PR #42, opened 2021-08-06
Add detailed error message for Provider processing

chadell:
Addressing #40 

TODO:

- [x] Better error message for providers `process()` method

### DIFF
--- a/circuit_maintenance_parser/parser.py
+++ b/circuit_maintenance_parser/parser.py
@@ -139,10 +139,10 @@ class ICal(Parser):
                     result.append(Maintenance(**data))
 
         except ValidationError as exc:
-            raise MissingMandatoryFields(exc)  # pylint: disable=W0707
+            raise MissingMandatoryFields from exc
 
         except Exception as exc:
-            raise ParsingError(exc)  # pylint: disable=W0707
+            raise ParsingError from exc
 
         logger.debug("Successful parsing for %s", self.__class__.__name__)
 
@@ -175,10 +175,10 @@ class Html(Parser):
             return result
 
         except ValidationError as exc:
-            raise MissingMandatoryFields(exc)  # pylint: disable=W0707
+            raise MissingMandatoryFields from exc
 
         except Exception as exc:
-            raise ParsingError(exc)  # pylint: disable=W0707
+            raise ParsingError from exc
 
     def parse_html(self, soup: ResultSet, data_base: Dict) -> Iterable[Union[Mapping[str, Union[str, int, Dict]]]]:
         """Custom HTML parsing."""
--- a/circuit_maintenance_parser/parser.py
+++ b/circuit_maintenance_parser/parser.py
@@ -139,10 +139,10 @@ class ICal(Parser):
                     result.append(Maintenance(**data))
 
         except ValidationError as exc:
-            raise MissingMandatoryFields from exc
+            raise MissingMandatoryFields(exc)  # pylint: disable=W0707
 
         except Exception as exc:
-            raise ParsingError from exc
+            raise ParsingError(exc)  # pylint: disable=W0707
 
         logger.debug("Successful parsing for %s", self.__class__.__name__)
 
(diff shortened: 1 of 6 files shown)


circuit_maintenance_parser/parser.py
```diff
@@ -175,10 +175,10 @@ def process(self) -> Iterable[Maintenance]:
             return result
 
         except ValidationError as exc:
-            raise MissingMandatoryFields from exc
+            raise MissingMandatoryFields(exc)  # pylint: disable=W0707
 
         except Exception as exc:
-            raise ParsingError from exc
+            raise ParsingError(exc)  # pylint: disable=W0707
```
glennmatthews, line 181: Can you explain the reason for this change? Pylint certainly seems to think it's a step in the wrong direction. If it's just to make it easier for logging purposes, note that `raise X from Y` will attach `Y` under `X.__cause__` so you can access it that way if desired.

circuit_maintenance_parser/providers.py
```diff
@@ -73,12 +76,15 @@ def process(self, data_type: str = "") -> Iterable[Maintenance]:
                 )
                 if data_type and data_type == parser.get_data_type() or not data_type:
                     return parser.process()
-            except (ParsingError, MissingMandatoryFields):
+            except (ParsingError, MissingMandatoryFields) as exc:
                 logger.debug(
-                    "Parser %s for provider %s was not successful", parser_class.__name__, self.__class__.__name__
+                    "Parser %s for provider %s was not successful: %s", parser_name, provider_name, exc,
```
glennmatthews, line 81: Given that this is a `debug` log, I'd personally prefer it to contain the full traceback, something like:

```python
logger.debug(
    "Parser %s for provider %s was not successful:\n%s",
    parser_name,
    provider_name,
    traceback.format_exc(),
)
```

circuit_maintenance_parser/providers.py
```diff
@@ -73,12 +76,15 @@ def process(self, data_type: str = "") -> Iterable[Maintenance]:
...
                 )
+                error_message += f"- Parser class {parser_name} from {provider_name} failed due: {exc}\n"
```
glennmatthews, line 83:
```suggestion
                error_message += f"- Parser class {parser_name} from {provider_name} failed due to: {exc}\n"
```

tests/unit/test_parsers.py
```diff
@@ -100,23 +100,60 @@ def test_complete_parsing(parser_class, raw_file, results_file):
...
+            ICal,
+            Path(dir_path, "data", "ical", "ical_no_account"),
+            MissingMandatoryFields,
+            "1 validation error for Maintenance\naccount\n  String is empty or 'None' (type=value_error)",
+        ),
+        (
+            ICal,
+            Path(dir_path, "data", "ical", "ical_no_maintenance_id"),
+            MissingMandatoryFields,
+            "1 validation error for Maintenance\nmaintenance_id\n  String is empty or 'None' (type=value_error)",
```
glennmatthews, line 116: For readability, consider using triple-quoted strings with literal newlines, e.g.:

```suggestion
            """\
1 validation error for Maintenance
maintenance_id
  String is empty or 'None' (type=value_error)""",
```

circuit_maintenance_parser/providers.py
```diff
@@ -73,12 +76,15 @@ def process(self, data_type: str = "") -> Iterable[Maintenance]:
...
                 continue
-        raise ParsingError("None of the parsers was able to parse the notification")
+        raise ParsingError(
+            f"None of the {provider_name} parsers was able to parse the notification.\nDetails:\n{error_message}"
+        )
```
glennmatthews, line 87: Would it be reasonable to additionally attach the list of actual exceptions as an additional attribute to the ParsingError, so that the caller has the option to introspect them if desired?

circuit_maintenance_parser/parser.py
chadell: Will double-check it

glennmatthews (approved):
Looks good!

glennmatthews (approved):

PR #46, opened 2021-08-12
Downgrade click version because conflicts with Celery

chadell:
found dependency issues when working with Celery

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ include = [
 
 [tool.poetry.dependencies]
 python = "^3.6.1"
-click = "^8.0"
+click = "^7.1"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -22,7 +22,7 @@ include = [
 
 [tool.poetry.dependencies]
 python = "^3.6.1"
-click = "^8.0"
+click = "^7.1"
```
glennmatthews, line 25: Might even consider explicitly allowing both, i.e. `>=7.1, <9.0`) so as to avoid encountering this issue in reverse if/when Celery moves to using 8.x?

glennmatthews (approved):

PR #49, opened 2021-08-16
Lumen Variations

carbonarok:
After looking at some email notifications I've spotted some varriations in the strings used to match data. There is also a possibility that a notification can have an empty table.

`"scheduled maintenance #:"` can sometimes be used to identify the ID, and `"This maintenance is scheduled"` can be used to find the status.

Testing has been added to ensure these variations.

### DIFF
--- a/circuit_maintenance_parser/parsers/lumen.py
+++ b/circuit_maintenance_parser/parsers/lumen.py
@@ -11,7 +11,6 @@ from circuit_maintenance_parser.parser import Html, Impact, CircuitImpact, Statu
 
 # pylint: disable=too-many-nested-blocks, too-many-branches
 
-import traceback
 
 logger = logging.getLogger(__name__)
 
@@ -28,7 +27,6 @@ class HtmlParserLumen1(Html):
             return [data]
 
         except Exception as exc:
-            traceback.print_exc()
             raise ParsingError from exc
 
     def parse_spans(self, spans: ResultSet, data: Dict):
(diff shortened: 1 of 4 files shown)


circuit_maintenance_parser/parsers/lumen.py
```diff
@@ -59,38 +65,39 @@ def parse_tables(self, tables: ResultSet, data: Dict):
...
-            if cells[0].string == "Start" and cells[1].string == "End":
-                num_columns = 2
-                for idx in range(num_columns, len(cells), num_columns):
-                    if "GMT" in cells[idx].string and "GMT" in cells[idx + 1].string:
-                        start = parser.parse(cells[idx].string.split(" GMT")[0])
-                        data["start"] = self.dt2ts(start)
-                        end = parser.parse(cells[idx + 1].string.split(" GMT")[0])
-                        data["end"] = self.dt2ts(end)
-                        break
+            if cells:
```
glennmatthews, line 68: Can we invert this logic to reduce nesting?

```python
            if not cells:
                continue
            if cells[0].string == "Start" and cells[1].string == "End":
                num_columns = 2
                ...
```

circuit_maintenance_parser/parsers/lumen.py
```diff
@@ -24,18 +25,20 @@ def parse_html(self, soup, data_base):
         try:
             self.parse_spans(soup.find_all("span"), data)
             self.parse_tables(soup.find_all("table"), data)
-
             return [data]
 
         except Exception as exc:
+            traceback.print_exc()
```
glennmatthews, line 31: Should this be reverted? Or is it an intentional addition?

carbonarok: Sorry, left over code. My bad

chadell (approved):
LGTM

PR #50, opened 2021-08-20
Add a Verizon parser

pke11y:
This PR adds an email parser for Verizon circuit maintenance notifications.

Unit/Provider tests included:
- Confirmed notification with 1 circuit
- Confirmed notifications with 4 circuits
- Cancelled notification with 2 circuits


circuit_maintenance_parser/__init__.py
```diff
@@ -15,6 +15,7 @@
     PacketFabric,
     Telia,
     Telstra,
+    Verizon,
```
chadell, line 18: could you also update Readme/Changelog?

circuit_maintenance_parser/parsers/verizon.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class HtmlParserVerizon1(Html):
+    """Notifications Parser for Verizon notifications."""
+
+    def parse_html(self, soup, data_base):
+        """Execute parsing."""
+        data = data_base.copy()
+        try:
+            logger.debug("PARSING")
```
chadell, line 23: parsing what? maybe a more detailed message would help :)

circuit_maintenance_parser/providers.py
```diff
@@ -152,6 +153,13 @@ class Telstra(GenericProvider):
     _default_organizer = "gpen@team.telstra.com"
 
 
+class Verizon(GenericProvider):
+    """Verizon provider custom class."""
+
+    _parser_classes: Iterable[Type[Parser]] = [HtmlParserVerizon1]
+    _default_organizer = "NO-REPLY-sched-maint@EMEA.verizonbusiness.com"
```
chadell, line 160: maybe not the most "generic" email, but until we don't have other, it should work

chadell (approved):
It looks good! just few simple points to complete it

circuit_maintenance_parser/providers.py
pke11y: agreed, but the only one we have right now.

circuit_maintenance_parser/parsers/verizon.py
pke11y: That was only temporary, but will keep and add more detail if that works. Can remove if required.

chadell (approved):
LGTM

circuit_maintenance_parser/parsers/verizon.py
```diff
@@ -0,0 +1,81 @@
...
+    @staticmethod
+    def parse_p(p_tags: ResultSet, data: Dict):
+        """Parse <p> tag."""
+        for p_tag in p_tags:
+            p_text = p_tag.text.strip()
+            if p_text.startswith("Dear"):
+                match = re.search(r"Dear (.*),", p_text)
+                if match:
+                    data["account"] = match.group(1)
+                    break
```
glennmatthews, line 81: Perhaps simpler?

```suggestion
            p_text = p_tag.text.strip()
            match = re.match(r"Dear (.*),", p_text)
            if match:
                data["account"] = match.group(1)
                break
```

glennmatthews (approved):

chadell (approved):
LGTM

PR #53, opened 2021-08-28
New library design

chadell:
# TODO

- [x] Simplify data to only one class with several conversion methods
- [x] Take into account that an email can contain multiple notifications (no only Ical), due to multiple payloads
- [x] Check for DRY
- [x] Add init data for Gmail API -> **NO NEEDED**, we can leverage on Gmail API message directly as it requires multiple calls to dowload attachements, etc., and the parsed don't handle connections, so what we will do is to use the [raw](https://developers.google.com/gmail/api/reference/rest/v1/users.messages) option to download the message as a regular email  an RFC 2822
- [x] Refactor testing
- [x] Document changes in changelog and readme

The main idea behind this refactor is to make the library more "autonomous" to get all the information that it needs to create notifications. The main change is that, before, a notifications was parsed only by ONE parser each time, now we could combine multiple parsers to get all the information needed.

The workflow would look like:
1. Init the `Provider` (if no provider type, a standard Icalendar one is offered by default). 

```python
from circuit_maintenance_parser import init_provider
ntt_provider = init_provider("ntt")
```

2. Init the `Data` to be parsed

```python
raw_data = b"""BEGIN:VCALENDAR
VERSION:2.0
...
END:VCALENDAR
"""
from circuit_maintenance_parser import init_data_raw

data_to_process =init_data_raw("ical", raw_data)
```

In this case we are using the simples `init`, but an `init_from_email` could be used to call the library using a standard Email representation. This way, the library is not closely tangled with a specific data source, and can be reused in multiple use cases, but at the same time offers all the data structure processing, so the library user don't need to take care of.

3. Process the `Data` using the `Provider` processor (that is composed by multiple combined parsers)

```python
parsed_notifications = ntt_provider.process_notifications(data_to_process)

print(parsed_notifications[0].to_json())
{
  "account": "137.035999173",
  "circuits": [
    {
      "circuit_id": "acme-widgets-as-a-service",
      "impact": "NO-IMPACT"
    },
...
}
```


circuit_maintenance_parser/__init__.py
```diff
@@ -81,72 +79,9 @@ def get_provider_class_from_sender(email_sender: str) -> Type[GenericProvider]:
...
-    else:
-        raise NonexistentParserError(
-            f"{provider_name} is not a currently supported provider. Only {', '.join(SUPPORTED_PROVIDER_NAMES)}"
-        )
-
-    return provider.get_data_types()
-
-
-# TODO: In Release 1.2.0, we introduced new Provider abstraction to accomodate multiple potential parser per provider.
-# In order to mantain backwards compatibility until a new major version release (2.x.x) we still mantain the following
```
glennmatthews, line 99: So just to confirm - after merging this PR, the next release will be 2.0.0 since these are breaking changes?

circuit_maintenance_parser/data.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+DataPart = namedtuple("DataPart", "type content")
+
+
+class NotificationData(BaseModel, extra=Extra.forbid):
+    """Base class for Notificaiton Data types."""
```
glennmatthews, line 18:
```suggestion
    """Base class for Notification Data types."""
```

glennmatthews, line 18: Can this docstring please be detailed a bit more? Maybe with examples of what is meant by "notification data", how this can store one or several DataParts, etc.?

circuit_maintenance_parser/data.py
```diff
@@ -0,0 +1,47 @@
...
+
+    _data_parts: List[DataPart] = []
+
+    def get_data_parts(self) -> List[DataPart]:
+        """Return an Iterable of Data objects from each NotificationData class."""
+        return self._data_parts
```
glennmatthews, line 24: Why does `_data_parts` need to be a private attribute with a dedicated getter function? Why not just a public instance attribute that can be accessed directly?

circuit_maintenance_parser/data.py
```diff
@@ -0,0 +1,47 @@
...
+        """Initialize the data_parts from an email format"""
+        for part in email_message.walk():
+            self._data_parts.append(DataPart(part.get_content_type(), part.get_payload()))
+
+        # Adding extra headers that are interesting to be parsed
+        self._data_parts.append(DataPart("email-header-subject", email_message["Subject"]))
+        # TODO: Date could be used to extend the "Stamp" time of a notification when not available, but we need a parser
+        self._data_parts.append(DataPart("email-header-date", email_message["Date"]))
+
+        return self
```
glennmatthews, line 47: Feels like these `init` functions should be `@classmethod`s, not instance methods, to me?

circuit_maintenance_parser/parser.py
```diff
@@ -23,42 +23,20 @@
...
-        Parser(raw=b'raw_message', default_provider='unknown', default_organizer='unknown')
-
-        >>> Parser(b"raw_message")
-        Traceback (most recent call last):
-        ...
-        TypeError: __init__() takes exactly 1 positional argument (2 given)
-
+    A Parser depend on a expected data type and depending on the implementation of the `run(raw)` method will
+    parse the `raw` data in order to identify some of the data that will eventually be used to create a
+    Maintenance.
```
glennmatthews, line 28:
```suggestion
    A Parser handles one or more specific data type(s) (specified in `_data_types`).
    The `parse(raw)` method must be implemented to parse the `raw` data to extract the
    (possibly partial/incomplete) data that will eventually be used to create a Maintenance object.
```

circuit_maintenance_parser/parser.py
```diff
@@ -23,42 +23,20 @@
...
-    _data_type = "text/plain"
+    # Data Type are used to match the type of parser to be used for each type of data
+    _data_types = ["text/plain", "plain"]
 
     @classmethod
-    def get_data_type(cls) -> str:
+    def get_data_types(cls) -> List[str]:
         """Return the expected data type."""
-        return cls._data_type
+        return cls._data_types
```
glennmatthews, line 37: As above - why does `_data_types` need to be private and hidden behind a getter function?

circuit_maintenance_parser/parser.py
```diff
@@ -23,42 +23,20 @@
...
 
-    def process(self) -> Iterable[Maintenance]:
+    def parse(self, raw: bytes) -> List[Dict]:
         """Method that returns a list of Maintenance objects."""
```
glennmatthews, line 40: Is this docstring (repeated below) correct now? Or should it be changed here and below to something like:

```suggestion
        """Extract a list of data that partially or completely describes a series of Maintenance objects."""
```

circuit_maintenance_parser/__init__.py
chadell: yes, this is the idea

circuit_maintenance_parser/data.py
chadell: sure, this PR is not finished (documentation, tests, etc.). I will take this into account after the initial review of the design

circuit_maintenance_parser/data.py
chadell: yes, true... this is coming from some initial step where I was hiding for some reason I don't remember :)

circuit_maintenance_parser/data.py
chadell: yes, they are implemented in a `self` way that is not matching the standard `factory` pattern, I will change it

circuit_maintenance_parser/parser.py
```diff
@@ -74,37 +52,31 @@ class ICal(Parser):
...
-                    organizer = (
-                        str(component.get("ORGANIZER")) if component.get("ORGANIZER") else self.default_organizer
-                    )
-                    provider = (
-                        str(component.get("X-MAINTNOTE-PROVIDER"))
-                        if component.get("X-MAINTNOTE-PROVIDER")
-                        else self.default_provider
-                    )
+                    organizer = str(component.get("ORGANIZER"))
+                    provider = str(component.get("X-MAINTNOTE-PROVIDER"))
```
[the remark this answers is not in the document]
chadell: they are actually used. When we were not able to get the `provider` and/or `organizer` info from the parsed notification we use some defaults per `provider`.
For example, if in the NTT notification this info is not present, we would use the defaults: `organizer = someemail@ntt.com` and `provider = cls.__name__`

circuit_maintenance_parser/processor.py
```diff
@@ -0,0 +1,121 @@
...
+from circuit_maintenance_parser.data import NotificationData
+from circuit_maintenance_parser.parser import Parser
+from circuit_maintenance_parser.errors import ParsingError, ProcessorError, ProviderError
+
+
+logger = logging.getLogger(__name__)
+
+
+class GenericProcessor(BaseModel, extra=Extra.forbid):
+    """Base class for the Processors."""
```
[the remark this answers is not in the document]
chadell: yes, sure

circuit_maintenance_parser/processor.py
```diff
@@ -0,0 +1,121 @@
...
+        self._post_process_hook()
+
+        return self.maintenances_data
+
+    def _process_hook(self, maintenances_extracted_data: List):
+        """Custom method per processor to accumulate the data from each DataPart."""
+        raise NotImplementedError
+
+    def _post_process_hook(self):
+        """Hook to add a post parsing process logic."""
```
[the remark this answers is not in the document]
chadell: because they are not supposed to be used by other classes than children classes, so I understand as internal all the "family". is this a wrong approach?

circuit_maintenance_parser/processor.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            raise ProcessorError(f"Unexpected data retrieved from parser: {maintenances_extracted_data}")
+
+    def post_process_hook(self, maintenances_data):
+        """After processing all the parsers, we try to combine all the data together."""
+        self.extend_processor_data(self.combined_maintenance_data)
+        try:
+            maintenances_data.append(Maintenance(**self.combined_maintenance_data))
+        except ValidationError as exc:
+            raise ProviderError("Not enough information available to create a Maintenance notification.") from exc
```
glennmatthews, line 119: I think I suggested `ProviderError` in a previous review comment, but actually, should this be `ProcessorError` now?

README.md
```diff
@@ -148,7 +145,7 @@ Circuit Maintenance Notification #0
 ```
 
 ```bash
-$ circuit-maintenance-parser --raw-file tests/integration/data/zayo/zayo1.html --parser zayo
+$ circuit-maintenance-parser --data-file tests/unit/data/zayo/zayo1.html --data-type html --provider-type zayo
```
glennmatthews, line 148: Just a friendly reminder that before we merge this, we probably want to also include a set of CHANGELOG updates spelling out what's changed between 1.x and 2.x, both in terms of CLI usage and integration as a library (e.g. into nautobot-circuit-maintenance).

circuit_maintenance_parser/data.py
```diff
@@ -0,0 +1,41 @@
+"""Definition of Data classes."""
+from collections import namedtuple
+from typing import List
+
+import email
+from pydantic import BaseModel, Extra
+
+
+DataPart = namedtuple("DataPart", "type content")
```
glennmatthews, line 9: Totally a nitpick or matter of style, but I find a list of fields more obvious/intuitive than a space-separated string of fields when defining a namedtuple:

```suggestion
DataPart = namedtuple("DataPart", ["type", "content"])
```

tests/unit/test_processors.py
```diff
@@ -0,0 +1,229 @@
...
+from circuit_maintenance_parser.parsers.telstra import HtmlParserTelstra1
+from circuit_maintenance_parser.parsers.verizon import HtmlParserVerizon1
+from circuit_maintenance_parser.parsers.zayo import HtmlParserZayo1
+
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+
+@pytest.mark.parametrize(
+    "processor_class, data_parsers, data_type, data_file, result_parse_file",
```
glennmatthews, line 27: Would it make more sense to pass in a list of `provider_class` and use each one's `processors` attribute in place of manually specifying combinations of `processor_class` and `data_parsers`? My concern here would be that this list could easily get out of sync with the actual provider/processor/parser combinations that we are actually implementing and caring about.

Alternately, is this test even needed, or is it functionally a duplicate of the testing in test_providers.py below?

chadell: sorry @glennmatthews , this is completely outadated, I'm refactoring the tests completely

circuit_maintenance_parser/parser.py
```diff
@@ -152,35 +120,29 @@ def process(self) -> Iterable[Maintenance]:
 class Html(Parser):
     """Html parser."""
 
-    _data_type = "text/html"
+    _data_types = ["text/html", "html"]
```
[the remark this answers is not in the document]
chadell: They are just arbitrary data types that you expect to match... so following your point, we could either document it or maybe create some ENUM that could help understanding them

circuit_maintenance_parser/parser.py
```diff
@@ -74,52 +52,45 @@ class ICal(Parser):
...
-                raise ParsingError from exc
+                raise ParserError from exc
 
         if not gcal:
-            raise ParsingError("Not a valid iCalendar data received")
+            raise ParserError("Not a valid iCalendar data received")
 
         try:
-            gcal = Calendar.from_ical(self.raw)
+            gcal = Calendar.from_ical(raw)
```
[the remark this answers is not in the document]
chadell: totally!

README.md
```diff
@@ -21,11 +21,15 @@ enables supporting other providers that are not using this proposed practice, ge
...
 
-Starting from a Provider parsing class, **multiple** parsers can be attached (in a specific order) and specific provider information (such as the default email used from the provider).
+1. We instantiate a `Provider`, directly or via the `init_provider` method, that depending on the selected type will return the corresponding instance.
+2. Each `Provider` have already defined multiple `Processors` that will be used to get the `Maintenances` when the `Provider.get_maintenances(data)` method is called.
+3. Each `Processor` class can have a pre defined logic to combine the data extracted from the notifications and create the final `Maintenance` object, and receives a `List` of multiple `Parsers` that will be to `parse` each type of data.
+4. Each `Parser` class supports multiple data types and implements the `Parser.parse()` method used to retrieve a `Dict` with the relevant key/values.
+5. When calling the `Provider.get_maintenances(data)`, the `data` argument is an instance of `NotificationData` that is just a collection of multiple `DataParts`, each one with a `type` and a `content` that will be used by the corresponding `Parser` when the `Processor` will try to match them.
```
glennmatthews, line 30:
```suggestion
5. When calling the `Provider.get_maintenances(data)`, the `data` argument is an instance of `NotificationData` (which is just a collection of multiple `DataParts`, each one with a `type` and a `content`) that will be used by the corresponding `Parser` when the `Processor` will try to match them.
```

glennmatthews (approved):
Honestly it's a bit hard to follow with all the refactoring and reorganization of code but I think this appears reasonable. Let's get it merged so that we can start actually trying out some new parsers with the combined data.

circuit_maintenance_parser/parser.py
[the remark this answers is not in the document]
pke11y: Makes sense. I'm sure documenting would suffice. No need to change type.

pke11y (dismissed):

PR #54, opened 2021-08-31
Turkcell Parser

carbonarok:
Added parser fo turkcell

### DIFF
--- a/circuit_maintenance_parser/parsers/turkcell.py
+++ b/circuit_maintenance_parser/parsers/turkcell.py
@@ -53,16 +53,17 @@ class HtmlParserTurkcell1(Html):
                         #   Eth-Trunk1.1               up      up       111111111111111|01-CUSTOMER|LOCATION|LINK
                         if re.match(r".+[ \t]([0-1]+\|.+\|.+\|.+)", element.text.strip()):
                             groups = re.search(r".+[ \t]([0-1]+\|.+\|.+\|.+)", element.text.strip())
-                            details = groups.group(1).split("|")
-                            data["circuits"].append(CircuitImpact(impact=Impact("OUTAGE"), circuit_id=details[0]))
-                            data["account"] = details[1]
+                            if groups:
+                                details = groups.group(1).split("|")
+                                data["circuits"].append(CircuitImpact(impact=Impact("OUTAGE"), circuit_id=details[0]))
+                                data["account"] = details[1]
 
         # Circuit table
         # Possibility that there could be a table inside the first table.
         if len(tables) == 2:
             tr_elements = tables[1].find_all("tr")
             data["circuits"] = []
-            for tr in tr_elements:
-                line = tr.text.strip().split("\n\n\n")
+            for tr_element in tr_elements:
+                line = tr_element.text.strip().split("\n\n\n")
                 data["circuits"].append(CircuitImpact(impact=Impact("OUTAGE"), circuit_id=line[0]))
                 data["account"] = line[1]
--- a/circuit_maintenance_parser/parsers/turkcell.py
+++ b/circuit_maintenance_parser/parsers/turkcell.py
@@ -1,0 +1,70 @@
+"""Turkcell parser."""
+import logging
+import re
+from typing import Dict
+
+from dateutil import parser
+import bs4  # type: ignore
+from bs4.element import ResultSet  # type: ignore
+
+from circuit_maintenance_parser.errors import ParsingError
+from circuit_maintenance_parser.parser import Html, Impact, CircuitImpact, Status
+
+# pylint: disable=too-many-nested-blocks, too-many-branches
+
+
+logger = logging.getLogger(__name__)
+
+
+class HtmlParserTurkcell1(Html):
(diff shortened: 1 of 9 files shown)


circuit_maintenance_parser/parsers/turkcell.py
```diff
@@ -0,0 +1,68 @@
...
+                    data["status"] = Status["CONFIRMED"]
+            if "Maintenance Number" in td_element.text.strip():
+                data["maintenance_id"] = td_elements[idx + 1].text.strip()
+            elif "Start" in td_element.text.strip():
+                data["start"] = self.dt2ts(parser.parse(td_elements[idx + 1].text.strip()))
+            elif "End" in td_element.text.strip():
+                data["end"] = self.dt2ts(parser.parse(td_elements[idx + 1].text.strip()))
+            elif "Impact of the maintenance" in td_element.text.strip():
+                data["summary"] = td_elements[idx + 1].span.text.strip()
+                if len(tables) == 1:
```
chadell, line 48: if I understand correctly, the Circuit info can be part of the "Impact of the maintenance" section or in a separate table?
maybe, to make this more evident, in this "elif" you could get the `p_elements` and then implement the if statement at the end:

```python
if len(tables) == 1 and p_elements:
...
elif len(tables) == 2:
...

```

circuit_maintenance_parser/parsers/turkcell.py
```diff
@@ -0,0 +1,69 @@
...
+    def parse_tables(self, tables: ResultSet, data: Dict):
+        """Parse tables."""
+        # Main table
+        td_elements = tables[0].find_all("td")
+        for idx, td_element in enumerate(td_elements):
+            if "Dear Customer" in td_element.text.strip():
+                if "planned" in td_element.text.strip():
+                    data["status"] = Status["CONFIRMED"]
+                else:
+                    data["status"] = Status["CONFIRMED"]
```
glennmatthews, line 39: Why have an if/else block that does the same thing in both branches? Is one of these Status values incorrect?

circuit_maintenance_parser/parsers/turkcell.py
```diff
@@ -0,0 +1,69 @@
...
+        data = data_base.copy()
+        try:
+            self.parse_tables(soup.find_all("table"), data)
+            return [data]
+
+        except Exception as exc:
+            raise ParsingError from exc
+
+    def parse_tables(self, tables: ResultSet, data: Dict):
+        """Parse tables."""
```
glennmatthews, line 31: I think as a general practice (even if we haven't been doing this previously) it would be a **really helpful thing** if in the docstring or a comment we were to include a stripped-down example of the HTML format we're expecting to find and parse, something like:

```suggestion
        """Parse tables.
        
        Format 1 supported here (all information in a single table):
            <table>
                <tr>
                    <td>Dear Customer ... planned ...</td>
                    <td>...</td>
                </tr><tr>
                    <td>Maintenance Number</td>
                    <td>123456789</td>
                </tr><tr>
                    <td>Start</td>
                    <td>2021-08-31 01:02:03</td>
                    ...
                    
        Format #2 supported here (one table for general information, a separate table listing impacted circuits):
            ...
        """
```

circuit_maintenance_parser/parsers/turkcell.py
```diff
@@ -0,0 +1,69 @@
...
+                data["summary"] = td_elements[idx + 1].span.text.strip()
+                if len(tables) == 1:
+                    data["circuits"] = []
+                    p_elements = td_elements[idx + 1].find_all("p")
+                    for element in p_elements:
+                        # Example match:
+                        #   Eth-Trunk1.1               up      up       111111111111111|01-CUSTOMER|LOCATION|LINK
+                        if re.match(r".+[ \t]([0-1]+\|.+\|.+\|.+)", element.text.strip()):
+                            groups = re.search(r".+[ \t]([0-1]+\|.+\|.+\|.+)", element.text.strip())
+                            if groups:
```
glennmatthews, line 56: Doesn't seem like the initial `re.match` call is needed - redundant?

```suggestion
                        groups = re.search(r".+[ \t]([0-1]+\|.+\|.+\|.+)", element.text.strip())
                        if groups:
```

circuit_maintenance_parser/parsers/turkcell.py
chadell: I understood that he only found the "planned" use case, and he was enforcing the default one, that is the same.
Ideally, we should have some `elif` in between when more use cases are added...
is this accurate Kristian?

circuit_maintenance_parser/parsers/turkcell.py
chadell: Yes, I agree that this would be really useful to provide context for future reviews.

circuit_maintenance_parser/parsers/turkcell.py
carbonarok: Yeah that is correct Christian. I need more examples with different status. Will be adding more in the future

glennmatthews (approved):

PR #55, opened 2021-09-03
Make Provider tests independent of particular Providers

chadell:
Provider tests should only test Provider methods logic, not particular Provider implementations, that are already tested in the End to End tests.


carbonarok (approved):
LGTM

PR #56, opened 2021-09-03
Extend email parsing support for encoded attachements

chadell:
While working on a new CSV parser, we found that we were not taking properly the payload data, so now we add the `decode=True` option

See: https://docs.python.org/3.5/library/email.message.html#email.message.Message.get_payload

### DIFF
--- a/circuit_maintenance_parser/data.py
+++ b/circuit_maintenance_parser/data.py
@@ -35,17 +35,31 @@ class NotificationData(BaseModel, extra=Extra.forbid):
         return cls.init_from_emailmessage(email_message)
 
     @classmethod
+    def walk_email(cls, email_message, data_parts):
+        """Recursive walk_email using Set to not duplicate data entries."""
+        for part in email_message.walk():
+            if "image" in part.get_content_type():
+                # Not interested in parsing images/QRs yet
+                continue
+
+            if "multipart" in part.get_content_type():
+                for inner_part in part.get_payload():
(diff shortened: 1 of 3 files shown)


circuit_maintenance_parser/data.py
```diff
@@ -35,17 +35,31 @@ def init_from_email_bytes(cls, raw_email_bytes: bytes):
...
+            elif "message/rfc822" in part.get_content_type():
+                if isinstance(part.get_payload(), email.message.Message):
+                    cls.walk_email(part.get_payload(), data_parts)
+            else:
+                data_parts.add(DataPart(part.get_content_type(), part.get_payload(decode=True)))
+
+    @classmethod
+    def init_from_emailmessage(cls, email_message):
+        """Initialize the data_parts from an email.message.Email object."""
+        data_parts = set()
```
glennmatthews, line 58: Why a set? Do we have some case where we get duplicate parts that we need to discard?

chadell: Yes, an email is full of duplicated payload, for instance from the EML file, the attachement is seen in 4 different parts! (and this happens for the rest of content).
So, as the example I tested, I was getting 24 parts, that after the SET were reduced to only 7.
I believe that it's better to use the recursive function and get as many things as possible and then prune for duplication that missing some piece of data.

pke11y (approved):
LGTM

PR #59, opened 2021-09-04
issue #59: Take `Stamp` from Email Date 

chadell:
Addresses #59 
* Create a new parser to take the `Stamp` from the Email data
* Use the Parser in most of the HTML based Providers, using `CombinedProcessor` instead of the `SimplerProcessor`


circuit_maintenance_parser/parser.py
```diff
@@ -161,3 +162,21 @@ def clean_line(line):
...
         return line.replace("=C2", "").replace("=A0", "").replace("\r", "").replace("=", "").replace("\n", "")
+
+
+class EmailDateParser(Parser):
+    """Parser for Email Date."""
+
+    _data_types = ["email-header-date"]
+
+    def parse(self, raw: bytes) -> List[Dict]:
+        """Method that returns a list of Maintenance objects."""
```
carbonarok, line 173: Does this method only support specific formats of dates?

carbonarok (approved):
LGTM

circuit_maintenance_parser/parser.py
chadell: Only supports Email Date format

Issue #62, opened 2021-09-05
Enable library client to define default `provider` and `organizer` for `Maintenances`


chadell:
### Proposed Functionality

Enable the client of the library to initalize the `Provider` with a default `provider` type and `organizer` to overwrite the default one that every `Provider` has.

### Use Case

This enable better customization, for instance, if for a specific client a network service provider uses a custom email or the client wants to use a custom type name.

chadell:
Because the output can be manipulated afterwards, it's not necessary to overcomplicate the default mechanism. Not gonna be implemented

PR #63, opened 2021-09-06
Issue #62: Enable custom provider type and organizer

chadell:
Addresses issue #62 
By default, the `Provider` classes provide some info to the `Maintenance` when this is not available from the data parsed.
Specifically:
* `provider`: that by default matches to the `Provider` class name in lower case.
* `organizer`: email address from the provider.

Both values can be inferred from the class, but now the library also offers two optional attributes: `provider_type` and `organizer` to enable the client to customize this data to different values if desired.


circuit_maintenance_parser/__init__.py
```diff
@@ -73,30 +75,29 @@ def get_provider_class(provider_name: str) -> Type[GenericProvider]:
     """Returns the Provider parser class for a specific provider_type."""
     provider_name = provider_name.lower()
 
-    for provider_parser in SUPPORTED_PROVIDERS:
-        if provider_parser.get_provider_type() == provider_name:
+    for provider in SUPPORTED_PROVIDERS:
+        if provider().get_provider_type() == provider_name:
```
glennmatthews, line 79: I'll admit I don't quite understand how this feature is supposed to work/be used. We have the option to override a provider's `provider_type` at init time, but when we're looking up a provider to initialize, we have to use its base (non-overridden) `provider_type`? What does the overridden value get used for, and what does the base value get used for - are there places where they're in conflict? Can I override a provider's `provider_type` to match that of another existing base provider class, (`init_provider(provider_type="ntt", extended_data={"provider_type": "packetfabric"})`) and does that totally break things if so?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
   - New `NotificationData` class that enables easier data injection for simple and complex data objects, such as
     emails.
   - Tests refactor to make them more specific to each type of data, mocking interfaces between different classes.
+- #63 - Enable library client to define default `provider_type` and `organizer` to be used to create `Maintenances`.
```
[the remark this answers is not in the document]
chadell: yes, it would. Now it was just the merging order, but we should sort it by number

circuit_maintenance_parser/__init__.py
chadell: As you say, the `provider_type` is used in the `init_provider`, to match the class name in the factory pattern.
Later, it's not used at all until the `Provider`get_extended_data` method is used to complement the `Maintenance` data.
The user case would be.
1. The client (Plugin) uses a provider named `centurylink`
2. Via its own mapping, he will use the `init_provider("lumen", extended_data={"provider_type":"centurylink"}`
3. This will return the proper Provider, but when we recieve the processed `Maintenances` the `provider` will be `centurylink` instead of the default `lumen`
Obviously, this is something that you could overwrite when the object is received, so not 100% sure if it's worth the change...

glennmatthews: Yeah, this is feeling like a solution in search of a problem to me... I'm not yet convinced this is necessary or desirable, seems like it'll just be a source of confusion.

circuit_maintenance_parser/__init__.py
```diff
@@ -38,19 +38,21 @@
     Zayo,
 )
 
-SUPPORTED_PROVIDER_NAMES = [provider.get_provider_type() for provider in SUPPORTED_PROVIDERS]
-SUPPORTED_ORGANIZER_EMAILS = [provider.get_default_organizer() for provider in SUPPORTED_PROVIDERS]
+SUPPORTED_PROVIDER_NAMES = [provider().get_provider_type() for provider in SUPPORTED_PROVIDERS]
```
pke11y, line 41: `SUPPORTED_PROVIDER_NAMES` seems a little bit confusing when the function that is populating the list is calling `get_provider_type`. Ultimately the function returns a class name but do we envisage type being extended in any way. If so,  `SUPPORTED_PROVIDER_TYPES` might make more sense.

PR #65, opened 2021-09-06
Adding new Lumen notification parser test

pke11y:
Adding a new test notification for Lumen.


pke11y:
Failing pytest as below. Not sure why just yet as quite similar to other tests. Enum evaluation seems to be different.

```
         +   'circuits': [CircuitImpact(circuit_id='1-12345678', impact=<Impact.OUTAGE: 'OUTAGE'>)],
         -   'circuits': [{'circuit_id': '1-12345678',
         -                 'impact': 'OUTAGE'}],
..
         +   'status': <Status.IN_PROCESS: 'IN-PROCESS'>},
         -   'status': 'IN-PROCESS',
```

chadell:
> Failing pytest as below. Not sure why just yet as quite similar to other tests. Enum evaluation seems to be different.
> 
> ```
>          +   'circuits': [CircuitImpact(circuit_id='1-12345678', impact=<Impact.OUTAGE: 'OUTAGE'>)],
>          -   'circuits': [{'circuit_id': '1-12345678',
>          -                 'impact': 'OUTAGE'}],
> ..
>          +   'status': <Status.IN_PROCESS: 'IN-PROCESS'>},
>          -   'status': 'IN-PROCESS',
> ```

maybe is this the only diff element? I've found that even complaining, the actual `assert` worked fine

chadell:
@pke11y I've run it locally, focus on:
```
E         -   'summary': 'Lumen will implement scheduled maintenance as part of a large '
E         -              'scale project to upgrade code on the IP platform. This '
E         -              'maintenance activity is one of several maintenances taking place '
E         -              'across various locations in multiple regions. The upgrade will '
E         -              'mitigate an existing defect in the current software version and '
E         -              'prevent future network outages.'},
```

pke11y:
@chadell yes, looking at that now thanks. Debug showed it's not picking up `summary` somehow. Investigating..

circuit_maintenance_parser/parsers/lumen.py
```diff
@@ -39,7 +39,10 @@ def parse_spans(self, spans: ResultSet, data: Dict):
                 ):
                     data["maintenance_id"] = line_text.split("#: ")[-1]
                 elif line_text.startswith("summary:"):
-                    for sibling in line.next_siblings:
+                    siblings = line.next_siblings
+                    if not line.next_sibling:
```
chadell, line 43:
```suggestion
                    if not siblings:
```

chadell (commented):
could you add it to `e2e_test`?

circuit_maintenance_parser/parsers/lumen.py
pke11y: Had that initially, but `siblings` is actually a generator. So even if it's empty it's not `None`. Cleanest solution was to use the bs4 lib, which returns `None` when no sibling.

chadell: sure, now I remember

tests/unit/test_e2e.py
```diff
@@ -85,6 +85,18 @@
                 Path(dir_path, "data", "date", "email_date_1_result.json"),
             ],
         ),
+        (
```
chadell, line 88: check new pattern:
```        (
            Lumen,
            [
                ("html", Path(dir_path, "data", "lumen", "lumen2.html")),
                ("email-header-date", Path(dir_path, "data", "date", "email_date_1")),
            ],
            [
                Path(dir_path, "data", "lumen", "lumen2_result.json"),
                Path(dir_path, "data", "date", "email_date_1_result.json"),
            ],
        ),
```
it takes into account the email date as stamp

chadell (approved):
LGTM

PR #66, opened 2021-09-06
Momentum Provider

carbonarok:
Added momentum parser using combined parsers. The status and the maintenance ID is held in the email subject.


tests/unit/test_e2e.py
```diff
@@ -78,6 +79,13 @@
             Path(dir_path, "data", "megaport", "megaport2.html"),
             Path(dir_path, "data", "megaport", "megaport2_result.json"),
         ),
+        # Momentum
+        (
+            Momentum,
+            "email",
```
chadell, line 85: merge from `develop`, this has changed a bit

carbonarok: will do that now

chadell (approved):
LGTM (we only need to merge `develop`)

chadell (approved):
LGTM

PR #67, opened 2021-09-07
Refactor Parser (DRY)

chadell:
* Implement a base skeleton `parse` method
* Add a `parser_hook` method that is the one implemented in each `Parser` class
* Raise `ParserError` when a `Parser` returns empty data


circuit_maintenance_parser/parser.py
```diff
@@ -37,10 +37,25 @@ def get_data_types(cls) -> List[str]:
         """Return the expected data type."""
         return cls._data_types
 
-    def parse(self, raw: bytes) -> List[Dict]:
-        """Extract a list of data that partially or completely describes a series of Maintenance objects."""
+    def parser_hook(self, raw: bytes):
```
glennmatthews, line 40: Is the new `parser_hook` expected to return a `List[Dict]`? If so it's probably still helpful to include the type hinting throughout.

circuit_maintenance_parser/parser.py
```diff
@@ -37,10 +37,25 @@ def get_data_types(cls) -> List[str]:
...
+        """Custom parser logic."""
         raise NotImplementedError
 
+    def parse(self, raw: bytes) -> List[Dict]:
+        """Execute parsing."""
+        try:
+            result = self.parser_hook(raw)
+            if result == [{}]:
```
glennmatthews, line 48: Kind of a weird special case, at a glance. Should we instead check for `result == []` or `(not item for item in result)` to check for the case where no data at all is returned and/or the case where multiple datas are returned but they're all empty?

circuit_maintenance_parser/parser.py
```diff
@@ -55,32 +70,19 @@ class ICal(Parser):
 
     _data_types = ["text/calendar", "ical", "icalendar"]
 
-    def parse(self, raw: bytes) -> List[Dict]:
-        """Method that returns a list of Maintenance objects."""
-        result = []
-
+    def parser_hook(self, raw: bytes):
```
glennmatthews, line 73: Can we document somewhere the conventions for the various `parse` methods and how they interrelate? We now have `parse`, which calls `parser_hook`, which can either call `parse_ical`, `parse_html` (which in turn might call `parse_body`, `parse_bs`, `parse_div`, `parse_p`, `parse_tables`, and/or `parse_title`), `parsedate_tz`, `parse_subject`, or `parse_csv` (is that all of them?)

circuit_maintenance_parser/parser.py
```diff
@@ -37,10 +37,25 @@ def get_data_types(cls) -> List[str]:
...
+            if result == [{}]:
+                raise ParserError(
+                    f"{self.__class__.__name__} has not extracted any valid data has been extracted from notification:"  # type: ignore
+                    f"\n{raw}"  # type: ignore
+                )
+            logger.debug("Successful parsing for %s", self.__class__.__name__)
+            return result
+
+        except Exception as exc:
+            raise ParserError from exc
```
glennmatthews, line 57: Note that the inner ParserError will be caught by the outer `try...except` block, which will wrap it in another ParserError... Perhaps instead?

```suggestion
        try:
            result = self.parser_hook(raw)
        except Exception as exc:
            raise ParserError from exc
        if result == [{}]:
            raise ParserError(
                f"{self.__class__.__name__} has not extracted any valid data has been extracted from notification:"  # type: ignore
                f"\n{raw}"  # type: ignore
            )
        logger.debug("Successful parsing for %s", self.__class__.__name__)
        return result
```

circuit_maintenance_parser/parser.py
chadell: Will do!

circuit_maintenance_parser/parser.py
```diff
@@ -37,24 +37,34 @@ def get_data_types(cls) -> List[str]:
...
+        `bs4.BeautifulSoup` and this is the method that each specific parser will implement to finally extract the
+        desired data from the raw content.
+        """
         raise NotImplementedError
 
     def parse(self, raw: bytes) -> List[Dict]:
-        """Execute parsing."""
+        """Execute parsing.
+
+        Each main `Parser` class will implement its own custom logic within the `parser_hook` method.
```
glennmatthews, line 55:
```suggestion
        Do not override this method!
        Instead, each main `Parser` class should implement its own custom logic within the `parser_hook` method.
```

circuit_maintenance_parser/parser.py
```diff
@@ -37,10 +37,35 @@ def get_data_types(cls) -> List[str]:
...
+        """
+        try:
+            result = self.parser_hook(raw)
+        except Exception as exc:
+            raise ParserError from exc
+        if any(not partial_result for partial_result in result):
+            raise ParserError(
+                f"{self.__class__.__name__} has not extracted any valid data has been extracted from notification:"  # type: ignore
+                f"\n{raw}"  # type: ignore
+            )
```
glennmatthews, line 65: Feel like this error should be rephrased, but I don't have exact wording to suggest. Maybe it'd be good to also include the full `result` value? As it's not obvious that this failure can occur if the parser returns a mixture of valid and empty results.

glennmatthews (approved):

Issue #58, opened 2021-09-04
Handle exceptions for `NotificationData.init*` methods


chadell:
### Environment

- Python version: 
- circuit_maintenance_parser version: 

### Expected Behavior

When a problem is detected initializing `NotificationData`, the user of the `init_data` methods should not get the exception

### Steps to Reproduce

1.
2.
3.

chadell:
Addressed in #72

PR #72, opened 2021-09-08
Issue-58: Ensure safety usage of the init_data methods

chadell:
Addresses issue #58 
Ensure `try/except` pattern for any issue found while initializing `NotificationData` from the library client, returning empty data.
I know it's too open, but because there could be multiple errors I tried to make it the simplest in the code instead of documenting all the potential exceptions each method could raise.

### DIFF
--- a/circuit_maintenance_parser/data.py
+++ b/circuit_maintenance_parser/data.py
@@ -1,6 +1,6 @@
 """Definition of Data classes."""
 import logging
-from typing import List, NamedTuple, Optional, TypeVar, Type, Set
+from typing import List, NamedTuple, Optional, Type, Set
 
 import email
 from pydantic import BaseModel, Extra
@@ -16,9 +16,6 @@ class DataPart(NamedTuple):
     content: bytes
 
 
-T = TypeVar("T", bound="NotificationData")  # pylint: disable=invalid-name
-
-
 class NotificationData(BaseModel, extra=Extra.forbid):
     """Base class for Notification Data types."""
 
@@ -29,7 +26,9 @@ class NotificationData(BaseModel, extra=Extra.forbid):
         self.data_parts.append(DataPart(data_type, data_content))
 
     @classmethod
-    def init_from_raw(cls: Type[T], data_type: str, data_content: bytes) -> Optional[T]:
+    def init_from_raw(
+        cls: Type["NotificationData"], data_type: str, data_content: bytes
+    ) -> Optional["NotificationData"]:
         """Initialize the data_parts with only one DataPart object."""
         try:
             return cls(data_parts=[DataPart(data_type, data_content)])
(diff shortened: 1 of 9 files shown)


glennmatthews (commented):
What's the purpose of these `init_data_*` methods? They seem like just really thin wrappers around `Notificationdata.init*` methods - why would I call these instead of calling those directly?

chadell:
> What's the purpose of these `init_data_*` methods? They seem like just really thin wrappers around `Notificationdata.init*` methods - why would I call these instead of calling those directly?

Yes, it is really "thinner" than my initial idea, but the main purpose is to expose a method to the client to avoid it know about the internal library details/classes, so, instead of using `NotificationData.init_from_email` your could directly call `init_from_email` and you don't need to now, if you don't want, about `NotificaitonData` class, but as you say the wrapper is so light that is not adding any other value while it's adding duplicated code/tests.
After considering it, understanding the benefit of the simple method as not worth enough, I would consolidate to the class factory.

tests/unit/test_data.py
```diff
@@ -9,15 +9,21 @@
...
+def test_init_from_raw():
+    """Test the init_data_raw function."""
+    data = NotificationData.init_from_raw("my_type", b"my_content")
     assert isinstance(data, NotificationData)
     assert len(data.data_parts) == 1
     assert data.data_parts[0].type == "my_type"
     assert data.data_parts[0].content == b"my_content"
 
 
+def test_init_data_raw_with_issue():
```
glennmatthews, line 21:
```suggestion
def test_init_from_raw_with_issue():
```

circuit_maintenance_parser/data.py
```diff
@@ -23,16 +29,24 @@ def add_data_part(self, data_type: str, data_content: bytes):
         self.data_parts.append(DataPart(data_type, data_content))
 
     @classmethod
-    def init(cls, data_type: str, data_content: bytes):
+    def init_from_raw(cls: Type[T], data_type: str, data_content: bytes) -> Optional[T]:
```
glennmatthews, line 32: I'm not real familiar with this TypeVar pattern - would the following be as correct?

```suggestion
    def init_from_raw(cls: Type["NotificationData"], data_type: str, data_content: bytes) -> Optional["NotificationData"]:
```

glennmatthews (approved):

circuit_maintenance_parser/data.py
chadell: Doing like this, `mypy` complains with `Name 'Notification' is not defined  [name-defined]`
With `TypeVar` you create the actual Type for something that is unbound yet.
Maybe there is a better way to do...

glennmatthews: Huh. I really thought forward-declaration via strings was supported (are you sure you used `"NotificationData"` and not `"Notification"`?). Not blocking in any case though, was just curious.

chadell: you were right!

glennmatthews (approved):

PR #73, opened 2021-09-09
Sparkle Provider

carbonarok:
New provider Sparkle has been added.

Because of the structure of the Sparkle notifications, a new feature has been impelemented to process multiple notifications from a single email. This meant modifications to the `CpmbinedProcessor` and the processor testing.


tests/unit/test_processors.py
```diff
@@ -14,8 +14,35 @@
 from circuit_maintenance_parser.parser import Parser
 
 
-PARSED_DATA = [{"a": "b"}, {"c": "d"}]
-EXTENDED_DATA = {"z": "x"}
+PARSED_DATA = [
```
chadell, line 17: I would keep the `fake` data style, to make this completely independent from th eactual values from the parsers

circuit_maintenance_parser/processor.py
```diff
@@ -117,16 +117,27 @@ class CombinedProcessor(GenericProcessor):
...
-            maintenances_data.append(Maintenance(**self.combined_maintenance_data))
-        except ValidationError as exc:
-            raise ProcessorError("Not enough information available to create a Maintenance notification.") from exc
+        if maintenances_data:
+            maintenances = maintenances_data.copy()
+            maintenances_data.clear()
+            for maintenance in maintenances:
+                try:
+                    combind_data = {**self.combined_maintenance_data, **maintenance}
+                    maintenances_data.append(Maintenance(**combind_data))
```
glennmatthews, line 134:
```suggestion
                    combined_data = {**self.combined_maintenance_data, **maintenance}
                    maintenances_data.append(Maintenance(**combined_data))
```

tests/unit/data/sparkle/sparkle1_html_parser_result.json
```diff
@@ -0,0 +1,44 @@
...
+                "circuit_id": "CIR0000000000001",
+                "impact": "OUTAGE"
+            }
+        ],
+        "end": 1628766000,
+        "maintenance_id": "33333",
+        "start": 1628737200,
+        "status": "COMPLETED",
+        "summary": "A scheduled work will be carried out to perform hardware upgrade in order in order to ensure the continued integrity of the network."
+    }
```
glennmatthews, line 43: Just to confirm I understand correctly  - for Sparkle, we are potentially getting multiple distinct maintenance_ids for the same circuit? Not multiple circuits for a single maintenance_id?

carbonarok: For the same circuit. The problem with this is that we only have one example to go off. There could be other variations of this.

circuit_maintenance_parser/processor.py
```diff
@@ -117,16 +117,27 @@ class CombinedProcessor(GenericProcessor):
 
     def process_hook(self, maintenances_extracted_data, maintenances_data):
         """All the parsers contribute with a subset of data that is extended."""
-        # We only expect one data object from these parsers
         if len(maintenances_extracted_data) == 1:
             self.combined_maintenance_data.update(maintenances_extracted_data[0])
         else:
-            raise ProcessorError(f"Unexpected data retrieved from parser: {maintenances_extracted_data}")
+            maintenances_data.extend(maintenances_extracted_data)
```
chadell, line 123: I believe it would be useful to document the new logic in the docstring.
To make clear that when the parser return one object we just accumulate it in the `combined_maintenance_data` but when it's multiple, we understand that there are multiple `maintenances` so we save them in the `maintenances_data` to be all extended at the end.
Maybe we could add a safety check if we try this `maintenances_data.extend` twice... this should only happen one time in this processor

circuit_maintenance_parser/processor.py
```diff
@@ -117,16 +117,27 @@ class CombinedProcessor(GenericProcessor):
...
+            maintenances_data.clear()
+            for maintenance in maintenances:
+                try:
+                    combind_data = {**self.combined_maintenance_data, **maintenance}
+                    maintenances_data.append(Maintenance(**combind_data))
+                except ValidationError as exc:
+                    raise ProcessorError(
+                        "Not enough information available to create a Maintenance notification."
+                    ) from exc
+        else:
```
chadell, line 139: what if you do something like:

```
if not maintenances_data:
    maintenance_data = [{}]

maintenances = maintenances_data.copy()
maintenances_data.clear()
...
```

so, you don't need the last try/except 😄

chadell (changes requested):
@carbonarok few points to be addressed

circuit_maintenance_parser/processor.py
carbonarok: Because your re-initalizing `maintenance_data` the value cannot be passed back out of the function.
So ive implemented the code before
``` python
        if not maintenances_data:
            maintenances = [{}]
        else:
            maintenances = maintenances_data.copy()
            maintenances_data.clear()
```

circuit_maintenance_parser/processor.py
```diff
@@ -117,16 +117,27 @@ class CombinedProcessor(GenericProcessor):
...
             self.combined_maintenance_data.update(maintenances_extracted_data[0])
         else:
-            raise ProcessorError(f"Unexpected data retrieved from parser: {maintenances_extracted_data}")
+            maintenances_data.extend(maintenances_extracted_data)
 
     def post_process_hook(self, maintenances_data):
-        """After processing all the parsers, we try to combine all the data together."""
+        """After processing all the parsers, we try to combine all the data together.
+
+        For some notifications there can be multiple maintenances in a single file. To handle this, maintenances are store in a
```
chadell, line 128: maybe move this explanation to `process_hook` where this logic actually take place?

circuit_maintenance_parser/parsers/sparkle.py
```diff
@@ -0,0 +1,94 @@
...
+    def parse_html(self, soup):
+        """Execute parsing."""
+        data = {}
+        try:
+            return self.parse_tables(soup.find_all("table"), data)
+        except Exception as exc:
+            raise ParserError from exc
+
+    def clean_string(self, string):
+        """Remove hex characters and new lines."""
```
glennmatthews, line 51: For my own education, can you point out in the test files some examples of hex characters that are needing to be removed? I'm not sure I understand what is meant by that.

tests/unit/test_processors.py
chadell: @carbonarok remember to address this too

circuit_maintenance_parser/parsers/sparkle.py
```diff
@@ -0,0 +1,94 @@
...
+                        data.append(ticket)
+                elif "circuits involved" in td_elements[0].text.lower():
+                    self.set_all_tickets(
+                        data,
+                        "circuits",
+                        [CircuitImpact(impact=Impact.OUTAGE, circuit_id=self.clean_line(td_elements[1].text))],
+                    )
+                elif "description of work" in td_elements[0].text.lower():
+                    self.set_all_tickets(data, "summary", self.clean_string(td_elements[1].text))
+        self.set_all_tickets(data, "status", Status.COMPLETED)
```
chadell, line 92: maybe `CONFIRMED`? `COMPLETED` means that the maintenance is over...

circuit_maintenance_parser/parsers/sparkle.py
carbonarok: Inside the sparkle test you can see strange charecters being used. Maybe they arent HEX, i can change the name if needed.
`1111111=C2=A0/ 22222=C2=A0/ 33333` is in the sparkle test where its really meant to be `1111111 / 22222 / 33333`

chadell (dismissed):
LGTM

circuit_maintenance_parser/parsers/sparkle.py
glennmatthews: Hm. I thought those were supposed to be removed by our use of `quopri` in the base HTML parser class. Are you seeing cases where these characters are still present in the processed HTML being passed through to the Sparkle parser?

carbonarok: Yeah, I think I saw this in another parser but unsure which as it was a while ago.

chadell (approved):
LGTM

glennmatthews (approved):

PR #83, opened 2021-09-24
Extend Readme

chadell:
Extend Readme with extra explanations


README.md
```diff
@@ -19,15 +19,32 @@ during a NANOG meeting that aimed to promote the usage of the iCalendar format.
 proposed iCalendar format, the parser is straight-forward and there is no need to define custom logic, but this library
 enables supporting other providers that are not using this proposed practice, getting the same outcome.
 
-You can leverage on this library in your automation framework to process circuit maintenance notifications, and use the standarised output to handle your received circuit maintenance notifications in a simple way.
+You can leverage on this library in your automation framework to process circuit maintenance notifications, and use the standarised [`Maintenace`](https://github.com/networktocode/circuit-maintenance-parser/blob/develop/circuit_maintenance_parser/output.py) to handle your received circuit maintenance notifications in a simple way. Every `maintenace` object contains, at least, the following attributes:
```
glennmatthews, line 22:
```suggestion
You can leverage this library in your automation framework to process circuit maintenance notifications, and use the standardized [`Maintenance`](https://github.com/networktocode/circuit-maintenance-parser/blob/develop/circuit_maintenance_parser/output.py) to handle your received circuit maintenance notifications in a simple way. Every `maintenance` object contains, at least, the following attributes:
```

README.md
```diff
@@ -19,15 +19,32 @@ during a NANOG meeting that aimed to promote the usage of the iCalendar format.
...
-2. Each `Provider` have already defined multiple `Processors` that will be used to get the `Maintenances` when the `Provider.get_maintenances(data)` method is called.
-3. Each `Processor` class can have a pre defined logic to combine the data extracted from the notifications and create the final `Maintenance` object, and receives a `List` of multiple `Parsers` that will be to `parse` each type of data.
-4. Each `Parser` class supports one or more data types and implements the `Parser.parse()` method used to retrieve a `Dict` with the relevant key/values.
-5. When calling the `Provider.get_maintenances(data)`, the `data` argument is an instance of `NotificationData` (which is just a collection of multiple `DataParts`, each one with a `type` and a `content`) that will be used by the corresponding `Parser` when the `Processor` will try to match them.
+2. Get an instance of the `NotificationData` class that groups together `DataParts` which contain some content and a specific type (that will match a specific `Parser`), and adding some factory methods to initialize it from a single content (as before) or directly from a raw email content or `email.message.EmailMessage` instance.
```
glennmatthews, line 39: This paragraph introduces a lot of terminology - tried to elaborate on it a bit to make it more digestible.

```suggestion
2. Get an instance of the `NotificationData` class. This instance groups together `DataParts` which each contain some content and a specific type (that will match a specific `Parser`). For example, a `NotificationData` might describe a received email message, with `DataParts` corresponding to the subject line and body of the email. There are factory methods to initialize a `NotificationData` describing a single chunk of binary data, as well as others to initialize one directly from a raw email message or `email.message.EmailMessage` instance.
```

README.md
```diff
@@ -19,15 +19,32 @@ during a NANOG meeting that aimed to promote the usage of the iCalendar format.
...
+3. Each `Provider` have already defined multiple `Processors` that will be used to get the `Maintenances` when the `Provider.get_maintenances(data)` method is called.
```
glennmatthews, line 40:
```suggestion
3. Each `Provider` uses one or more `Processors` that will be used to build `Maintenances` when the `Provider.get_maintenances(data)` method is called.
```

README.md
```diff
@@ -19,15 +19,32 @@ during a NANOG meeting that aimed to promote the usage of the iCalendar format.
...
+4. Each `Processor` class can have a custom logic to combine the data extracted from the notifications and create the final `Maintenance` object, and receives a `List` of multiple `Parsers` that will be to `parse` each type of data.
```
glennmatthews, line 41:
```suggestion
4. Each `Processor` class uses one or more `Parsers` to process each type of data that it handles. It can have custom logic to combine the parsed data from multiple `Parsers` to create the final `Maintenance` object.
```

README.md
```diff
@@ -19,15 +19,32 @@ during a NANOG meeting that aimed to promote the usage of the iCalendar format.
...
+5. Each `Parser` class supports one or more data types and implements the `Parser.parse()` method used to retrieve a `Dict` with the relevant key/values.
```
glennmatthews, line 42:
```suggestion
5. Each `Parser` class supports one or a set of related data types, and implements the `Parser.parse()` method used to retrieve a `Dict` with the relevant keys/values.
```

README.md
```diff
@@ -19,15 +19,32 @@ during a NANOG meeting that aimed to promote the usage of the iCalendar format.
...
+6. When calling the `Provider.get_maintenances(data)`, the `data` argument is an instance of `NotificationData` that will be used by the corresponding `Parser` when the `Processor` will try to match them.
```
glennmatthews, line 43: Maybe delete item 6? It isn't a separate step so much as a restatement of items 2 and 3 above.

README.md
```diff
@@ -65,15 +82,39 @@ By default, there is a `GenericProvider` that support a `SimpleProcessor` using
 The library is available as a Python package in pypi and can be installed with pip:
 `pip install circuit-maintenance-parser`
 
-## Usage
+## How to use it?
+
+The library requires two things:
+
+- The `notificationdata`: this is the data that the library will check to extract the maintenance notifications. It can be simple (only one data type and content) or from an email, with multiple data parts.
```
glennmatthews, line 89:
```suggestion
- The `notificationdata`: this is the data that the library will check to extract the maintenance notifications. It can be simple (only one data type and content, such as an iCalendar notification) or more complex (with multiple data parts of different types, such as from an email).
```

README.md
```diff
@@ -65,15 +82,39 @@ By default, there is a `GenericProvider` that support a `SimpleProcessor` using
...
+- The `provider` type: used to select the proper `Provider` which contains the `processor` logic to take the proper `Parsers` and use the data that they extract. By default, the `GenericProvider`(used when no other provider type is defined) will support parsing of `iCalendar` notifications using the recommended format.
```
glennmatthews, line 90:
```suggestion
- The `provider` identifier: used to select the proper `Provider` which contains the `processor` logic to take the proper `Parsers` and use the data that they extract. By default, the `GenericProvider` (used when no other provider type is defined) will support parsing of `iCalendar` notifications using the recommended format.
```

README.md
```diff
@@ -97,93 +138,63 @@ END:VEVENT
...
+"stamp": 1444435800,
+"start": 1444464000,
+"status": "TENTATIVE",
+"summary": "Maint Note Example",
+"uid": "42"
 }
 ```
 
-## CLI
+Notice that, either with the `GenericProvider` or `NTT` provider, we get the same result from the same data, because they are using exactly the same `Processor` and `Parser`. The only difference is that `NTT` notifications come without `organizer` and `provider` in the notification, and this info is fulfilled with some default values for the `Provider`, but in this case the original notification contains all the necessary information, so the defaults are not used.
```
glennmatthews, line 178: Maybe not a great example here? If NTT is the same as GenericProvider, why would I as a user ever want to use the NTT provider specifically? What purpose does it serve?

README.md
```diff
@@ -203,6 +214,68 @@ Circuit Maintenance Notification #0
 }
 ```
 
+## How to Extend the Library?
+
+Even the library aims to include support for as many providers as possible, it's likely that not all the thousands of NSP are supported and you may need to add support for some new one. Adding a new `Provider` is quite straight forward, and in the following example we are adding support for an imaginary ABCDE provider, that uses HTML notifications.
```
glennmatthews, line 219:
```suggestion
Even though the library aims to include support for as many providers as possible, it's likely that not all the thousands of NSP are supported and you may need to add support for some new one. Adding a new `Provider` is quite straightforward, and in the following example we are adding support for an imaginary provider, ABCDE, that uses HTML notifications.
```

README.md
```diff
@@ -203,6 +214,68 @@ Circuit Maintenance Notification #0
...
+First step is creating a new file: `circuit_maintenance_parser/parsers/abcde.py`. This file will contain all the custom parsers needed for the provider and it will import the base classes for each parser type from `circuit_maintenance_parser.parser`. In the example, we only need to import `Html` and in the child class implement the methods required by the class, in this case `parse_html()` which will return a `dict` with all the data that this `Parser` can extract. In this case we have to helper methods, `_parse_bs` and `_parse_tables` that implement the logic to navigate the notification data.
+
+```python
+from typing import Dict
+import bs4  # type: ignore
+from bs4.element import ResultSet  # type: ignore
+from circuit_maintenance_parser.parser import Html
+
+class HtmlParserABCDE1(Html):
+    def parse_html(self, soup) -> Dict:
```
glennmatthews, line 230: Probably would be good to document the type of `soup`, as users may not necessarily have prior experience with BS4.

README.md
chadell: I used this example to show that some Providers (such as NTT, PacketFabric,) are using the default iCalendar parser and only add some custom default data (_default_organizer) to extend notifications data that maybe don't contain all the mandatory attributes.

glennmatthews (dismissed):

glennmatthews (approved):

PR #84, opened 2021-09-27
AWS Direct Parser

carbonarok:
Parser for AWS notifications


circuit_maintenance_parser/parser.py
```diff
@@ -226,3 +226,24 @@ def parser_hook(self, raw: bytes):
...
+
+class Text(Parser):
+    """Html parser."""
+
+    _data_types = ["text/plain"]
+
+    def parser_hook(self, raw: bytes):
+        """Execute parsing."""
+        result = []
+        soup = bs4.BeautifulSoup(quopri.decodestring(raw), features="lxml")
```
chadell, line 239: does this look like a Html parser? maybe we should add the `text/plain` into that data_type. I see it too much HTML specific for a generic Text parser

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -0,0 +1,75 @@
...
+                search = re.search(
+                    r"([A-Z][a-z]{2}, [0-9]{1,2} [A-Z][a-z]{2,9} [0-9]{4} [0-9]{2}:[0-9]{2}:[0-9]{2} [A-Z]{2,3}) to ([A-Z][a-z]{2}, [0-9]{1,2} [A-Z][a-z]{2,9} [0-9]{4} [0-9]{2}:[0-9]{2}:[0-9]{2} [A-Z]{2,3})",
+                    line,
+                )
+                if search:
+                    data["start"] = self.dt2ts(parser.parse(search.group(1)))
+                    maintenace_id += str(data["start"])
+                    data["end"] = self.dt2ts(parser.parse(search.group(2)))
+                if "may become unavailable" in line.lower():
+                    impact = "DEGRADED"
```
chadell, line 69: unavailable = OUTAGE

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -0,0 +1,75 @@
...
+            elif re.match(r"[a-z]{5}-[a-z0-9]{8}", line):
+                maintenace_id += line
+                data["circuits"].append(CircuitImpact(circuit_id=line, impact=Impact(impact)))
+        data["maintenance_id"] = hashlib.md5(maintenace_id.encode("utf-8")).hexdigest()  # nosec
```
chadell, line 73: this is a hot point, I have to think about it

circuit_maintenance_parser/provider.py
```diff
@@ -116,6 +117,15 @@ class AquaComms(GenericProvider):
     _default_organizer = "tickets@aquacomms.com"
 
 
+class AWS(GenericProvider):
+    """AWS provider custom class."""
+
+    _processors: List[GenericProcessor] = [
+        CombinedProcessor(data_parsers=[EmailDateParser, TextParserAWS1, SubjectParserAWS1]),
+    ]
+    _default_organizer = "aaaaaaaaaaaaaaa"
```
chadell, line 126: I guess this is wrong

tests/unit/data/aws/aws1_result.json
```diff
@@ -0,0 +1,41 @@
...
+                "impact": "DEGRADED"
+            },
+            {
+                "circuit_id": "aaaaa-00000006",
+                "impact": "DEGRADED"
+            }
+        ],
+        "end": 1621519200,
+        "maintenance_id": "21264c558a1b6a2bedaae0878d6318b0",
+        "organizer": "aaaaaaaaaaaaaaa",
```
chadell, line 32: wrong

circuit_maintenance_parser/parsers/aws.py
chadell: Thinking on this point, maybe we could get the maintenance id with the `start_time`, `end_time` and `services`.
So far, I'm not aware of any "re-schedule" from AWS notifications, and the cancellation ones are matching the same hash, so we could handle the state machine for Circuit Maintenance

circuit_maintenance_parser/parser.py
```diff
@@ -226,3 +226,23 @@ def parser_hook(self, raw: bytes):
...
+        # Even we have not noticed any HTML notification with more than one maintenance yet, we define the
```
chadell, line 240: this is not true :) , we have Sparkle

circuit_maintenance_parser/parser.py
```diff
@@ -226,3 +226,23 @@ def parser_hook(self, raw: bytes):
     def parse_csv(raw: bytes) -> List[Dict]:
         """Custom CSV parsing."""
         raise NotImplementedError
+
+
+class Text(Parser):
+    """Html parser."""
```
chadell, line 232: ?

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -0,0 +1,80 @@
...
+                )
+                if search:
+                    data["start"] = self.dt2ts(parser.parse(search.group(1)))
+                    data["end"] = self.dt2ts(parser.parse(search.group(2)))
+                    maintenace_id += str(data["start"])
+                    maintenace_id += str(data["end"])
+                if "may become unavailable" in line.lower():
+                    impact = Impact.OUTAGE
+                elif "has been cancelled" in line.lower():
+                    impact = Impact.NO_IMPACT
```
chadell, line 73: I would not change the impact type, even Cancelled

circuit_maintenance_parser/parser.py
```diff
@@ -226,3 +226,26 @@ def parser_hook(self, raw: bytes):
...
+    def parser_hook(self, raw: bytes):
+        """Execute parsing."""
+        result = []
+        text = self.get_text_hook(raw)
+        for data in self.parse_text(text):
+            result.append(data)
+        return result
+
+    @staticmethod
+    def get_text_hook(raw):
```
chadell, line 245:
```suggestion
    def get_text_hook(raw: bytes) -> str:
```

chadell (dismissed):
LGTM (update changelog and add a note for the `maintenance_id` hash point for future reference)

chadell (approved):
Awesome work @carbonarok !

PR #86, opened 2021-09-27
Ensure that CombinedProcessor reinitializes self.combined_maintenance_data

### LINKED ISSUE #85: CombinedProcessor can carry over data from previous parsing
### Environment

- Python version: 
- circuit_maintenance_parser version: 2.0.1

### Expected Behavior

A Provider that uses a CombinedProcessor should be able to cleanly process multiple NotificationData in series without data bleeding over from one processing to the next

### Observed Behavior

After successfully processing one NotificationData, then processing a second, incomplete NotificationData, the second processing returns a Maintenance that retains fields from the first processed NotificationData instead of raising a ProviderError due to the incomplete data.

### Steps to Reproduce

1.
2.
3.

glennmatthews:
Fixes #85 

The root cause was that for CombinedProcessor, `self.combined_maintenance_data` was only initialized at class instantiation time, meaning that when processing a second set of data using the same processor, any keys previously set on  `self.combined_maintenance_data` would bleed over from one processing to the next. The solution is to reinitialize this dictionary each time `process()` is called.

### DIFF
--- a/circuit_maintenance_parser/processor.py
+++ b/circuit_maintenance_parser/processor.py
@@ -115,6 +115,10 @@ class CombinedProcessor(GenericProcessor):
     # The CombinedProcessor will consolidate all the parsed data into this variable
     combined_maintenance_data: Dict = {}
 
+    def process(self, data: NotificationData, extended_data: Dict) -> Iterable[Maintenance]:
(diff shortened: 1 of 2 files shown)


chadell (approved):
LGTM

PR #90, opened 2021-09-28
Handle more potential Lumen statuses seen in the field

glennmatthews:
Handle the potential Lumen status values of "Not Completed" and "Postponed", as seen in actual emails.

Additionally, for some updates to an existing maintenance, there is no column in the table describing the status. To avoid an error in this case (status is a mandatory field), I'm setting the status to a default of "CONFIRMED" - is this the best way to handle this?


chadell (approved):
LGTM, but if we had some test sample could be even better :)

glennmatthews:
The samples I have fully pass existing parsing except for the differing status values, so I wasn't sure they were worth adding as additional test cases. I can redact them and add them as tests if you think it's worthwhile.

Issue #82, opened 2021-09-22
Be able to handle gracefully notifications that do not contain any notification info


chadell:
### Environment

- circuit_maintenance_parser version: 2.0.1

### Proposed Functionality

Today, a `Provider` processing a notification can only return either the maintenance objects or raise a `ProviderError` when it was not possible to get, at least, a maintenance object from the notification.
Some times the notification parsed is not actually containing any relevant information and it's not actually needed to be parsed and raise an error, just signal that it was not containing any relevant info.

### Use Case

Being able to differentiate between parsing errors and non relevant notification would make the library clients able to handle each one in a different way.

glennmatthews:
An example of the need for this is with Lumen, where emails regarding support tickets come from the same address (`no-reply@lumen.com`) as emails regarding scheduled maintenance. This results in a lot of "failed" parsing for the support ticket emails.

chadell:
Fixed in #91

PR #91, opened 2021-09-30
Provider implements a include filter to define relevant notifications

chadell:
Addressing #82 

The goal is to make the `Provider` capable of filtering in (and implicitly out), which are the relevant notifications that must be parsed and if not successful a `ProviderError` must be raised.

With the `_include_filter` dictionary, the `Provider` can define a dictionary of different data_type to match some string in, and if the filter exists, and the notification doens't match, instead of raising error, returns an empty list of `Maintenances`

## TODO, once approach is approved:

- [x] Add tests
- [x] Update readme


chadell:
@glennmatthews , take this as a proposal, it's just a draft, I will add tests and improve it (I believe it's time to make the keywords data type standard), but just raising to get early feedback on the overall idea

circuit_maintenance_parser/provider.py
```diff
@@ -59,12 +69,44 @@ class GenericProvider(BaseModel):
...
+    def filter_check(filter_dict: Dict, data: NotificationData) -> bool:
+        """Generic filter check."""
+        for data_part in data.data_parts:
+            filter_data_type = data_part.type
+            if filter_data_type not in filter_dict:
+                continue
+
+            data_part_content = data_part.content.decode()
+            if any(re.search(filter_re, data_part_content) for filter_re in filter_dict[filter_data_type]):
+                return True
```
glennmatthews, line 97: Would probably be helpful to log a debug message when a match is found, to aid in troubleshooting as to why a particular message was or wasn't filtered.

circuit_maintenance_parser/provider.py
```diff
@@ -59,12 +69,44 @@ class GenericProvider(BaseModel):
...
+        return False
+
     def get_maintenances(self, data: NotificationData) -> Iterable[Maintenance]:
         """Main entry method that will use the defined `_processors` in order to extract the `Maintenances` from data."""
         provider_name = self.__class__.__name__
         error_message = ""
         related_exceptions = []
 
+        if self.exclude_filter_check(data) or not self.include_filter_check(data):
+            return []
```
glennmatthews, line 108: Similarly, logging a message here would probably help.

circuit_maintenance_parser/provider.py
```diff
@@ -59,12 +69,44 @@ class GenericProvider(BaseModel):
...
+    def exclude_filter_check(self, data: NotificationData) -> bool:
+        """If `_exclude_filter` is defined, it verifies that the matching criteria is met."""
+        if self._exclude_filter:
+            return self.filter_check(self._exclude_filter, data)
+        return False
+
+    @staticmethod
+    def filter_check(filter_dict: Dict, data: NotificationData) -> bool:
+        """Generic filter check."""
+        for data_part in data.data_parts:
```
glennmatthews, line 90: I wonder - should we log a warning or something if we have a data type in `filter_dict` that isn't present in `data`? I.e. if we include based on `email-header-subject` but the data we're processing is missing that field?

circuit_maintenance_parser/provider.py
```diff
@@ -172,6 +214,8 @@ class HGC(GenericProvider):
 class Lumen(GenericProvider):
     """Lumen provider custom class."""
 
+    _include_filter = {EMAIL_HEADER_SUBJECT: ["Scheduled Maintenance Window"]}
```
glennmatthews, line 217: Looking at more examples that I have locally, I think just "Scheduled Maintenance" would be better here. I see some subject lines like `Lumen Scheduled Maintenance #: 22194642, Scheduled` that appear to me to be valid maintenance notifications.

circuit_maintenance_parser/provider.py
chadell: yes, I was not logging enough, will improve :)

circuit_maintenance_parser/provider.py
```diff
@@ -49,6 +50,10 @@ class GenericProvider(BaseModel):
             that will be used. Default: `[SimpleProcessor(data_parsers=[ICal])]`.
         _default_organizer (optional): Defines a default `organizer`, an email address, to be used to create a
             `Maintenance` in absence of the information in the original notification.
+        _include_filter (optional): Dictionary that defines matching string per data type to take a notification into
+            account.
+        _exclude_filter (optional): Dictionary that defines matching string per data type to NOT take a notification
+            into account.
```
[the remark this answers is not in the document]
chadell: yes, you are right, I am making `exclude` to take precedence. I've added a note

circuit_maintenance_parser/provider.py
```diff
@@ -50,6 +52,14 @@ class GenericProvider(BaseModel):
             that will be used. Default: `[SimpleProcessor(data_parsers=[ICal])]`.
         _default_organizer (optional): Defines a default `organizer`, an email address, to be used to create a
             `Maintenance` in absence of the information in the original notification.
+        _include_filter (optional): Dictionary that defines matching regex per data type to take a notification into
+            account.
+        _exclude_filter (optional): Dictionary that defines matching regex per data type to NOT take a notification
+            into account.
+
+    Notes:
+        - If a notification matches both, the `_include_filter` and `_exclude_filter`, the second takes precedence and
```
glennmatthews, line 61:
```suggestion
        - If a notification matches both the `_include_filter` and `_exclude_filter`, the exclusion takes precedence and
```

glennmatthews (approved):

PR #102, opened 2021-10-26
Add parsers for Equinix maintenance notifications via email

scetron:


glennmatthews:
Looks like there are still some `mypy` linting failures for the new code.

circuit_maintenance_parser/parsers/equinix.py
```diff
@@ -0,0 +1,116 @@
...
+            impact (Status object): impact of the maintenance notification (used in the parse table function to assign an impact for each circuit).
+        """
+        for b_elem in b_elements:
+            if "UTC:" in b_elem:
+                raw_time = b_elem.next_sibling
+                # for non english equinix notifications
+                # english section is usually at the bottom
+                # this skips the non english line at the top
+                if not self._isascii(raw_time):
+                    continue
```
glennmatthews, line 63: This feels pretty fragile to me but we can go with it for now and fix later if it proves to be an issue.

glennmatthews (approved):

circuit_maintenance_parser/parsers/equinix.py
scetron: Agreed. Unfortunately seemed to be the best way to detect the non-english characters which don't encode to ascii. If they stop including English, this would break, but than we'd have a larger problem of localization.

PR #104, opened 2021-10-30
Use local locations DB

chadell:
Use a 41K locations CSV file to resolve location coordinates locally if possible before trying external API queries if not present.

Other enhancements:
* Using `backoff` library to handle retries to external DB 
* Using `pandas`(latest `python3.6` compatible version) to handle the local CSV DB
* Refactor the geolocation into a class `Geolocator` that initializes the local DB and the timezone resolver once for all.


circuit_maintenance_parser/utils.py
```diff
@@ -1,27 +1,116 @@
...
+        return location_coordinates
+
+    def get_location_from_local_file(self, city: str) -> Tuple[float, float]:
+        """Get location from Local DB."""
+        city_name = city.split(", ")[0]
+        country = city.split(", ")[-1]
+
+        if self.db_location is not None:
+            # 1. We try to match city name and country
+            location = self.db_location.query(f"city_ascii == '{city_name}' & country == '{country}'")
```
glennmatthews, line 64: Is pandas overkill for this problem? I wonder if we could instead do something relatively simple with the built-in `csv` library, something like:

```python
self.db_location = {}
with open('worldcities.csv') as csvfile:
    reader = csv.DictReader(csvfile)
    for row in reader:
        # Index by city and country
        self.db_location[(row["city_ascii"], row["country"])] = (row["lat"], row["lng"])
        # Index by city (first entry wins if duplicated names)
        if row["city_ascii"] not in db_location:
            self.db_location[row["city_ascii"]] = (row["lat"], row["lng"])

...

lat, lng = self.db_location.get((city_name, country), self.db_location.get(city_name, None))

File: circuit_maintenance_parser/utils.py [hunk 2]
```diff
@@ -1,27 +1,116 @@
...
+            "Resolved city %s to coordinates: lat %s - lon %s", city, location_coordinates[0], location_coordinates[1],
+        )
+        return location_coordinates
+
+    def get_location_from_local_file(self, city: str) -> Tuple[float, float]:
+        """Get location from Local DB."""
+        city_name = city.split(", ")[0]
+        country = city.split(", ")[-1]
+
+        if self.db_location is not None:
```
glennmatthews, line 62: Do we need this check? It doesn't look like `load_db_location()` has any case where it'd fail to populate `self.db_location`.

circuit_maintenance_parser/utils.py
```diff
@@ -1,27 +1,116 @@
...
+        logger.debug("Resolved %s to %s from OpenStreetMap webservice.", city, location)
+        return (location.latitude, location.longitude)
+
+    def city_timezone(self, city: str) -> str:
+        """Get the timezone for a given city.
+
+        Args:
+            city (str): Geographic location name
+        """
+        if self.timezone is not None:
```
glennmatthews, line 95: Is this check needed? Same reason.

circuit_maintenance_parser/utils.py
```diff
@@ -1,27 +1,116 @@
...
+            try:
+                latitude, longitude = self.get_location(city)
+                timezone = self.timezone.tzNameAt(latitude, longitude)
+                if not timezone:
+                    # If even with the location we can't find the timezone, we try getting the location
+                    # from API.
+                    latitude, longitude = self.get_location_from_api(city)
+                    timezone = self.timezone.tzNameAt(latitude, longitude)
```
glennmatthews, line 103: Shouldn't be needed since `get_location` already falls through on its own to calling `get_location_from_api` if `get_location_from_local_file` fails?

tests/unit/test_utils.py
```diff
@@ -2,26 +2,28 @@
...
+        ("Atlanta, GA", "America/New_York"),
+        ("New York, NY", "America/New_York"),
+        ("Worcester, MA", "America/New_York"),
+        ("Salt Lake City, Utah", "America/Denver"),
+        ("Los Angeles, CA", "America/Los_Angeles"),
+        ("Dublin, Ireland", "Europe/Dublin"),
+        ("Barcelona, Spain", "Europe/Madrid"),
+        ("Sydney, Australia", "Australia/Sydney"),
+        ("Guadalajara, Spain", "Europe/Madrid"),
+        ("Guadalajara, Mexico", "America/Mexico_City"),
```
glennmatthews, line 24: Just confirming - are these a mix of "city + country matches CSV" and "city alone matches CSV" so that we are testing both cases?

circuit_maintenance_parser/utils.py
chadell: yes, it may be. I will try this implementation and if the time penalty is not significant, I also prefer to get rid of this dependency.

circuit_maintenance_parser/parsers/cogent.py
```diff
@@ -17,6 +17,8 @@
 class HtmlParserCogent1(Html):
     """Notifications Parser for Cogent notifications."""
 
+    _geolocator = Geolocator()
```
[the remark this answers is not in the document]
chadell: totally, good point. Even  for now it's good, it's better to be prepared for reusability.

circuit_maintenance_parser/utils.py
chadell: typing check 😓

circuit_maintenance_parser/utils.py
chadell: same as before, maybe it assumes that the method is available for class, no instance and fails the validation

circuit_maintenance_parser/utils.py
chadell: in theory yes, but I found a case, with the original DB data, that was resolving to Sydney coordinates and the timezone resolution was returning `None`, maybe due some bug on that library, because with the coordinates from the API was working well (some decimals of difference)

tests/unit/test_utils.py
chadell: exactly, this test was to validate that case, we try to get the city + country first to avoid ambiguities

circuit_maintenance_parser/utils.py
pke11y: Yes, I'd second that. Ideally we could use `DictReader` in the `csv` module. The `pandas` lib is quite heavy with more dependencies.

pke11y (approved):
Nice one!

glennmatthews (approved):

PR #109, opened 2021-11-11
Extend Zayo parser to handle additional examples

glennmatthews:
Handle some additional Zayo email notifications that were failing the current parser.

I'm marking this as a draft for now as I have some questions about how to handle information that may be omitted from certain Zayo notification emails (see TODO comments in zayo.py).


circuit_maintenance_parser/parsers/zayo.py
```diff
@@ -32,10 +42,23 @@ def parse_bs(self, btags: ResultSet, data: dict):
...
                     data["maintenance_id"] = self.clean_line(line.next_sibling)
-                elif line.text.lower().strip().startswith("urgency:"):
-                    urgency = self.clean_line(line.next_sibling)
-                    if urgency == "Planned":
+                elif "serves as official notification" in line.text.lower():
+                    if "will be performing maintenance" in line.text.lower():
                         data["status"] = Status("CONFIRMED")
+                    elif "has cancelled" in line.text.lower():
+                        data["status"] = Status("CANCELLED")
+                # TODO: some Zayo notifications may include multiple activity dates. How best to handle this?
```
retnuh, line 50: I'm guessing this may be one of the first instances where one email could theoretically generate multiple maintenances?  I'm guessing it would be a big change to change that cardinality all the way through...

I wonder if we could leverage the tagging feature in the other open PR to say that an email needs to get re-processed for a 2nd/3rd/nth date?  That could also be tricky though in terms of tracking the state of which activity needs to get re-processed.

circuit_maintenance_parser/parsers/zayo.py
```diff
@@ -50,9 +73,13 @@ def parse_bs(self, btags: ResultSet, data: dict):
                             break
                 elif line.text.lower().strip().startswith("reason for maintenance:"):
                     data["summary"] = self.clean_line(line.next_sibling)
+                # TODO: not all Zayo notifications include "date notice sent".
+                # Do we need to pull the date from the email headers, or what should we do here?
```
chadell, line 77: if the info is not consistent, I would defer to the `Date` header from email that is always there

circuit_maintenance_parser/parsers/zayo.py
```diff
@@ -50,9 +73,13 @@ def parse_bs(self, btags: ResultSet, data: dict):
...
                 elif line.text.lower().strip().startswith("date notice sent:"):
                     stamp = parser.parse(self.clean_line(line.next_sibling))
                     data["stamp"] = self.dt2ts(stamp)
+                # TODO: not all Zayo notifications include "customer".
+                # Do we need to pull the customer name from the email subject line, or what should we do here?
```
chadell, line 82: if the info of the customer is available in the subject, then we could use it adding a new parser to the processor

circuit_maintenance_parser/parsers/zayo.py
chadell: The constraint I see here is that the Maintenance ID is unique for the 3 periods, so I see it as one Maintenance with a non continuous time window. It's not an easy problem to solve because it is not matching the "[standard](https://github.com/jda/maintnote-std/blob/master/standard.md)" that this library is following. More specific, in this [section](https://github.com/jda/maintnote-std/blob/master/standard.md#35-event-component-and-associated-properties) the DTSTART and DTEND are unique for each notification.

Brainstorming some options:

- Get the overall maintenance time, taking the first START date and the last END date. From the example: `01-Nov-2021 06:01 - 03-Nov-2021 11:00`. This is the easiest one to implement and the issue is that there is lost of information for the time slots not affected.
- Create different Maintenances from the original one with different Maintenance IDs adding some suffix or prefix, for example from TTN-0001 we get TTN-0001-a, TTN-0001-b, TTN-0001-c. This is still keeping the standard like the one above (the parser supports taking multiple maintenances from a notification), but the client could get some misunderstanding from this.
- Create different Maintenances from the original one with same Maintenance ID, but with some difference, for instance using the `UID` filter to differentiate them, so each one would have UID=1, UID=2 and UID=3. This could mean some changes on the parser clients if they want to handle them as unique entities.

Open to feedback

glennmatthews: I'm thinking that the first option (treat it as one long extended maintenance) is probably the simplest to implement and is the most conservative option in terms of blocking an overall window where the circuit may be affected.

circuit_maintenance_parser/data.py
```diff
@@ -77,7 +77,8 @@ def init_from_emailmessage(cls: Type["NotificationData"], email_message) -> Opti
             # Adding extra headers that are interesting to be parsed
             data_parts.add(DataPart(EMAIL_HEADER_SUBJECT, email_message["Subject"].encode()))
             data_parts.add(DataPart(EMAIL_HEADER_DATE, email_message["Date"].encode()))
-            return cls(data_parts=list(data_parts))
+            # Ensure the data parts are processed in a consistent order
+            return cls(data_parts=sorted(data_parts, key=lambda part: part.type))
```
chadell, line 81: why order is important?

circuit_maintenance_parser/parser.py
```diff
@@ -171,11 +171,9 @@ def parse_html(self, soup: ResultSet,) -> List[Dict]:
     def clean_line(line):
         """Clean up of undesired characters from Html."""
         try:
-            line = line.text.strip()
+            return line.text.strip()
         except AttributeError:
-            line = line.strip()
-        # TODO: below may not be needed if we use `quopri.decodestring()` on the initial email file?
-        return line.replace("=C2", "").replace("=A0", "").replace("\r", "").replace("=", "").replace("\n", "")
+            return line.strip()
```
chadell, line 176: good cleanup

circuit_maintenance_parser/parsers/zayo.py
```diff
@@ -24,6 +43,14 @@ def parse_html(self, soup):
         self.parse_bs(soup.find_all("b"), data)
         self.parse_tables(soup.find_all("table"), data)
 
+        if data:
```
chadell, line 46: just wondering why do we need this check?

chadell (approved):
LGTM

circuit_maintenance_parser/data.py
glennmatthews: In the case where we have information of different accuracy/precision available from different data parts. For example with Zayo we can always pull an (approximate) `stamp` from the email date, but for some notifications we may have a more precise `stamp` explicitly stated in the email body. Without this change it was unpredictable as to which one of these would be processed last and therefore win out - see the CI failures on 7970ecc above for an example of this.

circuit_maintenance_parser/parsers/zayo.py
glennmatthews: For the case where we might decide that there's no relevant content in the email (and hence return an empty `data` dict) I wouldn't want to just fill in these additional fields in an otherwise empty dict.

chadell: I see, otherwise you could just rely on Pydantic validation once the `Maintenance` output is instantiated.

circuit_maintenance_parser/data.py
chadell: I understand, good catch.

PR #110, opened 2021-11-16
Support additional possible statuses for Telstra notifications

glennmatthews:
Add handling for and examples of Telstra notifications for statuses other than CONFIRMED.


chadell (approved):
LGTM

PR #111, opened 2021-11-17
Improve EXA (GTT) parser

glennmatthews:
Add examples of "rescheduled", "new", and "completed" EXA notifications and parser support for these.

### DIFF
--- a/circuit_maintenance_parser/parsers/gtt.py
+++ b/circuit_maintenance_parser/parsers/gtt.py
@@ -33,7 +33,9 @@ class HtmlParserGTT1(Html):
                     if groups:
                         data["maintenance_id"] = groups.groups()[0]
                         status = groups.groups()[1]
-                        if status == "Reminder":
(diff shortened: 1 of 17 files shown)


circuit_maintenance_parser/parsers/gtt.py
```diff
@@ -33,21 +33,39 @@ def parse_tables(self, tables, data):
                     if groups:
                         data["maintenance_id"] = groups.groups()[0]
                         status = groups.groups()[1]
-                        if status == "Reminder":
+                        if status == "New":
+                            data["status"] = Status["TENTATIVE"]
```
chadell, line 37: I would say that "New" would mean "CONFIRMED"?

glennmatthews: Yeah, I was on the fence about this one. I can change it.

chadell (approved):
LGTM

PR #112, opened 2021-11-17
Equinix improvements

glennmatthews:
- Avoid an undefined `impact` variable if the impact type is unrecognized (will likely still fail, but at least now we return a `None` rather than raising an exception)
- Add recognition of reduced-redundancy as an impact type
- Make the maintenance-id matching in the email subject line more conservative
- Add an `_include_filter` to ignore other types of Equinix emails
- Rename `equinix3_result.json` to `equinix1_result_combined.json` to more closely match its meaning and usage


chadell (approved):
LGTM

chadell (approved):
LGTM

PR #116, opened 2021-11-24
Colt parsing updates 

pke11y:
This PR will add a new `EmailSubject` parser for Colt to account for cancellation notifications. 

In addition, the existing `SubjectParserColt1` will be updated to accommodate carriage returns in the regex used to extract the data from the subject.

A test email was created with `Content-Transfer-Encoding: 7bit` as seen in production to test the parsing of the cancellation notice.


chadell (commented):
this looks great! I would just suggest to add a specific test for the new subject parser in `tests/unit/test_parsers.py`

chadell (approved):
LGTM

PR #120, opened 2021-11-29
Handle yet more Zayo variants

glennmatthews:
- Handle Zayo email subject lines with a non-standard number of `*` characters and/or missing content
- Ignore Zayo emails that don't include a "Maintenance Ticket #" in the email body.
- Handle a variant Zayo email about a completed circuit maintenance window.


chadell (approved):
LGTM

PR #122, opened 2021-11-30
Defer loading of tzwhere until it's needed

### LINKED ISSUE #121: tzwhere.tzwhere() is called at import time, causing substantial memory overhead
### Environment

- Python version: 
- circuit_maintenance_parser version: 2.0.4 and later

### Expected Behavior

No significant increase in memory consumption when loading the `circuit_maintenance_parser` library after upgrading from 2.0.3 to 2.0.4

### Observed Behavior

Memory usage jumps by over 400 MB after `import circuit_maintenance_parser` with 2.0.4.

This appears to be due to #104, specifically due to a `Geolocator` object being instantiated at library load time, which in turn calls `tzwhere.tzwhere()`, which appears to allocate a large amount of memory.

### Steps to Reproduce

1. `python`
2. `import circuit_maintenance_parser`

glennmatthews:
Fixes #121.

Instead of calling `tzwhere.tzwhere()` as a consequence of simply doing an `import circuit_maintenance_parser`, only call it when it's actually needed, by implementing it as a `classproperty` of the `Geolocator` class. Verified that this substantially reduces the baseline memory cost of `import circuit_maintenance_parser`.

I also explored replacing `tzwhere` with `timezonefinder`, which appears to be a more up-to-date, more memory-efficient fork, but found that while the peak memory cost of using `timezonefinder.TimezoneFinder()` instead of `tzwhere.tzwhere()` is much lower, `import timezonefinder` appears to cost about 20 MB more RAM than `import tzwhere`. So for the time being I didn't make that change, but we can revisit it if desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - #120 - Improve handling of Zayo notifications.
+- #121 - Defer loading of `tzwhere` data until it's needed, to reduce memory overhead.
 
 ## v2.0.6 - 2021-11-30
 
--- a/circuit_maintenance_parser/utils.py
+++ b/circuit_maintenance_parser/utils.py
@@ -16,36 +16,46 @@ logger = logging.getLogger(__name__)
 dirname = os.path.dirname(__file__)
 
 
+class classproperty:  # pylint: disable=invalid-name,too-few-public-methods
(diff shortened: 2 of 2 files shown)


chadell (approved):
Much better! LGTM

PR #125, opened 2021-12-03
Make 'status' optional, accept iCal notifications with no 'sequence' either

### LINKED ISSUE #115: How should iCal notifications with no X-MAINTNOTE-STATUS be handled?
### Environment

- circuit_maintenance_parser version: 2.0.5

### Proposed Functionality

Per the BCOP, `X-MAINTNOTE-STATUS` can be omitted from an iCal notification (the only example I see of this in the BCOP is for a rescheduled maintenance window), but our `Maintenance` class defines  `status` as a mandatory field. In general how should we handle iCal notifications that omit this field?

### Use Case

I have a real-world example received from a provider for a "completed" notification that omits the X-MAINTNOTE-STATUS, so simply assuming that "no X-MAINTNOTE-STATUS implies "status=re-scheduled"" would not be correct in this case. We could possibly parse the `DESCRIPTION` or `SUMMARY` fields -- in this particular case the word "completed" is present in both -- but that's probably not desirable as a general solution? Looking for suggestions here.

glennmatthews:
Fixes #115.

The real-world iCal notification that served as the basis of #115 turns out to also have no SEQUENCE field. Unlike the omitted X-MAINTNOTE-STATUS field, omitting SEQUENCE is not valid per the BCOP standard, but we have to make do with what we see in the real world, sigh. Let me know if you disagree with the way I'm handling the omission of either or both of these fields.


chadell (dismissed):
LGTM, but I would update the readme too to make this clear about the exceptions we are doing.

pke11y (approved):
Doc updates look good.

PR #126, opened 2021-12-03
Add exclude filter for Telia

glennmatthews:
Exclude customer support emails from Telia that don't include a maintenance notification.


chadell (approved):
LGTM

PR #127, opened 2021-12-07
More fixes

glennmatthews:
- Some Equinix emails don't clearly state "SCHEDULED" or "COMPLETED" (etc.) in their subject - default status of these emails to `CONFIRMED` for lack of any better contextual information.
- Lumen parser wasn't handling status `Cancelled` - add support for it.
- I think I went a bit too far in #125 by defaulting `status = Status.NO_CHANGE` globally, it's better to keep it as something that will raise an error if the parser doesn't set an appropriate value for it.


chadell (approved):
LGTM, could you update the Readme with the change on `status`?

chadell (approved):
LGTM

pke11y (approved):
Looks good

PR #141, opened 2022-05-10
Release v2.0.9

chadell:


pyproject.toml
```diff
@@ -33,22 +33,21 @@ tzwhere = "^3.0.3"
...
-black = "^19.10b0"
-pylint = "^2.4.4"
-pydocstyle = "^6.0.0"
-yamllint = "^1.20.0"
-bandit = "^1.6.2"
-invoke = "^1.4.1"
-toml = "0.10.2"
-flake8 = "^4.0.0"
-mypy = "^0.910"
+pytest = "*"
```
glennmatthews, line 36: Can you remind me why we consider it preferable to have wildcard versions for dev dependencies in `pyproject.toml`? It makes it so that inspection of the much-less-human-readable `poetry.lock`is necessary to determine what versions of these dependencies this project is currently developed/tested against, which seems like a loss of useful information to me.

tests/unit/test_e2e.py
```diff
@@ -44,23 +44,43 @@
...
-        (GenericProvider, [("ical", GENERIC_ICAL_DATA_PATH),], [GENERIC_ICAL_RESULT_PATH,],),
+        (
+            GenericProvider,
+            [
+                ("ical", GENERIC_ICAL_DATA_PATH),
+            ],
+            [
+                GENERIC_ICAL_RESULT_PATH,
+            ],
+        ),
```
glennmatthews, line 55: This is a net loss in readability, IMHO. Can we just strip out some trailing commas to make the newer version of Black happy with the one-liners instead, e.g.:

```suggestion
        (GenericProvider, [("ical", GENERIC_ICAL_DATA_PATH)], [GENERIC_ICAL_RESULT_PATH]),
```

or:

```suggestion
        (
            GenericProvider,
            [("ical", GENERIC_ICAL_DATA_PATH)],
            [GENERIC_ICAL_RESULT_PATH],
        ),
```

here and below?

tests/unit/test_providers.py
```diff
@@ -71,7 +73,7 @@ class ProviderWithIncludeFilter(GenericProvider):
 
     # With a non matching data to include, the notification will be skipped and just return empty
     other_fake_data = NotificationData.init_from_raw("other type", b"other data")
-    assert ProviderWithIncludeFilter().get_maintenances(other_fake_data) == []
+    assert not ProviderWithIncludeFilter().get_maintenances(other_fake_data)
```
glennmatthews, line 76: Not sure I see why this is a desirable change - now the test will pass if `get_maintenances()` returns *any* falsey value such as `None`, `0`, etc.

pyproject.toml
chadell: We agreed it in dev-standards, and IIRC, the main point was to get, for testing, the latest version of the packages for linting and testing. This is mentioned in issue our internal repo, issue 111

tests/unit/test_e2e.py
chadell: `black` is adding these commas

tests/unit/test_providers.py
chadell: I agree on it. I twas a change suggested by the linter as equivalent, but it isn't

pyproject.toml
glennmatthews: My read on that was that it was done for cookiecutters so that *newly created* projects would automatically get the latest available. `poetry.lock` ensures that the version remains pinned thereafter to some exact version, so I don't see any advantage to unpinning dependencies in an existing project.

scetron (approved):
lgtm

chadell:
Going to create a new one

PR #143, opened 2022-05-11
Fix Equinix Parser

### LINKED ISSUE #142: Equinix Parser Tests Failing
### Environment

- Python version: 3.8.13
- circuit_maintenance_parser version: 2.0.8

### Expected Behavior

All tests to pass

### Observed Behavior

Equinix Parser test fails. The parsed time stamp appears to be different than the expected timestamp.

### Steps to Reproduce

1. Run pytest
2. Observe failure and diff
3.

scetron:
Fixes #142 .

The parser was failing due to the exclusion of the year of the maintenance in the timestamp. The detailed time from the notification does not include the year, which is listed on a different line.

This fix parses the start and end year and adds it to the start and end date/time to get an accurate timestamp.

### DIFF
--- a/circuit_maintenance_parser/parsers/equinix.py
+++ b/circuit_maintenance_parser/parsers/equinix.py
@@ -54,7 +54,15 @@ class HtmlParserEquinix(Html):
             impact (Status object): impact of the maintenance notification (used in the parse table function to assign an impact for each circuit).
         """
         impact = None
+        start_year = 0
+        end_year = 0
         for b_elem in b_elements:
+            if "SPAN:" in b_elem.text:
+                # Formated in DAY MONTH YEAR
+                # *SPAN: 02-JUL-2021 - 03-JUL-2021*
+                raw_year_span = b_elem.text.strip().split()
+                start_year = raw_year_span[1].split('-')[-1]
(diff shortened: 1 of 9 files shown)


circuit_maintenance_parser/parsers/equinix.py
```diff
@@ -54,18 +54,29 @@ def _parse_b(self, b_elements, data):
             impact (Status object): impact of the maintenance notification (used in the parse table function to assign an impact for each circuit).
         """
         impact = None
+        start_year = 0
+        end_year = 0
         for b_elem in b_elements:
+            if "SPAN:" in b_elem.text:
```
chadell, line 60: why `.text` is needed here and not in the other if statements?

scetron: the relevant text is in the text of the b element rather than next to the b element:

`<b>SPAN: 02-JUL-2021 - 03-JUL-2021</b>`

vs

`<b>UTC:</b> FRIDAY, 02 JUL 10:00 - FRIDAY, 02 JUL 15:00`

chadell (approved):
LGTM

PR #158, opened 2022-08-12
fix: update class SubjectParserColt1  search

### LINKED ISSUE #157: Proposed Update to Colt.py
### Environment

- Python version: <3.9>
- circuit_maintenance_parser version: <2.0.3>

### Expected Behavior
Colt email subject `Colt Service Affecting Maintenance Notification - CRQ1-xxxxx [20/8/2022 20:00:00 GMT - 21/8/2022 01:00:00 GMT] for YYYYY, XYZ` is not properly handle by the parser.

Colt email doesn't start with `[ EXTERNAL ]`, this field could have been add by the email platform used during the first tests. 

The regex in SubjectParserColt1 and SubjectParserColt2 should be modify to not enforce the presence of brackets at the beginning of the email.

### Observed Behavior
Colt Parser is expecting COLT email to start with `[ .* ]`, the 2 brackets should be see as facultative as well.

BeArchiTek:
### Issue fixed
Fix #157

### Context
Update the parser in colt.py to keep [ xx ] as facultative.


itdependsnetworks:
I think this will require some tests, can you add some here: https://github.com/networktocode/circuit-maintenance-parser/tree/develop/tests/unit/data/colt

BeArchiTek:
👋 @itdependsnetworks ,

Add colt6 for new test, it's colt5 without the `[ EXTERNAL ]` (got some issue anonymising the real email ...)

chadell:
It looks like there is some issues with the `summary` (also with the status in the new example)

BeArchiTek:
you are right @chadell , should be the same. Let's close this one

PR #163, opened 2022-09-06
Add BSO Provider

fdypua:


chadell (commented):
Thanks for your contribution!
Two notes, remember that the E2E test require a complete Maintenance data, so you would need to add some extra info from the email for the `stamp` and also retrieve the account.
Check the Cogent E2E composition below yours to see the pattern.
And, also remember to udpate the Readme with the new supported Provider!
BTW, you can run your tests locally with `inv tests --local`

circuit_maintenance_parser/parsers/bso.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def parse_html(self, soup):
+        """Execute parsing."""
+        data = {}
+        self.parse_tables(soup.find_all("table"), data)
+        return [data]
+
+    def parse_tables(self, tables: ResultSet, data: Dict):
+        """Parse Table tag."""
+        data["account"] = "Customer"  # No Account Name information in html/email details
```
chadell, line 26: I would add a more clear information such as:
```suggestion
        data["account"] = "Customer info unavailable"  # No Account Name information in html/email details
```

fdypua: Thank you so much for suggesting. I would love to update if there are any other errors.

chadell: only two small points:
* you should update the test results with the new "account" ("Customer info unavailable")
* if you could add typing to some parser methods missing, it would be great

chadell (approved):
LGTM

PR #165, opened 2022-09-17
[WIP] Add a CI check to validate IP address present in tests, and a helper to clean them

chadell:
TODO

- [ ] Improve regex for IPv4
- [ ] Add IPv6 regex


tests/unit/data/aquacomms/aquacomms1.eml
```diff
@@ -1,91 +1,91 @@
...
-      </tr>
-      <tr>
-        <td valign=3D"TOP" align=3D"LEFT"><font face=3D"Arial, Verdana, Hel=
-vetica, sans-serif" size=3D"2">Aqua Comms NOC</font></td>
-      </tr>
-    </tbody></table>
-  </font>
-
---000000000000015b7605ca799cf2--
+MIME-Version: 1.0
```
glennmatthews, line 1: Why are the entire files showing as diffs? Did we change newline formats or encoding or something?

tests/unit/data/telstra/telstra2_result.json
```diff
@@ -11,6 +11,6 @@
...
+        "summary": "Telstra C2C Segment 6 (Singapore - Philippines) Submarine Cable Repair. This is to notify that Telstra will be conducting 2x Repeater R36 & R41A Replacement Work on C2C Segment 6 (Singapore - Philippines due to suspected fauly Laser Diodes to reinstate the End-to-End connectivity of Faulty and Degraded Fiber Pairs and optimize the maximum performance of all Fiber Pairs on this Segment between Singapore and Philippines. Unprotected circuits will be down throughout the repair window. Schedule of this work may change as they are dependent on many outside factors (weather / rough sea / permits/ hostile waters etc) which are beyond Telstra control. Our sincere apologies for any inconvenience caused.\n\nUpdate #1: Change of schedule  due to availability of  cable repair ship CS Teneo"
```
glennmatthews, line 14: Hmm, okay - is the change from `\r\n` to `\n` throughout intended?

chadell: with the rewrite via `sys.stdout.write`, these `\r` are getting removed. I guess they were taken when getting the value from the screen, but are not relevant for the test. Do you think necessary to find the right adjustment in the replace mechanism to honour them?

itdependsnetworks:
Just a friendly thought when I think of anonymizing IPs and parsing

* When you are parsing, there can be differences introduced by changing the octet size, such as a table 
* When viewing the data, sometimes IPs are related and anonimzing them all to the same, it makes it difficult to correlate 

Does it make sense to consider swapping the first subnet as the following:
* 1.x.x.x for single digit first subnets
* 10.x.x.x for double digit first subnets
* 172.x.x.x for triple digit first subnets

and keep everything else the same?

tests/unit/data/aquacomms/aquacomms1.eml
glennmatthews: Looks like this is default behavior of `open()`, which `fileinput.input()` uses under the hood:

> When reading input from the stream, if `newline` is None, universal newlines mode is enabled. Lines in the input can end in '\n', '\r', or '\r\n', and these are translated into '\n' before being returned to the caller.

It's probably desirable in the long run; it just makes it hard to tell from the diffs for this specific PR exactly what's been changed.

chadell:
> and keep everything else the same?

The parsing should work the same for any kind of addresses, so if changing the octet size fails, it would be great to catch the error earlier.
So far, in all the test data sets, I have found only one IP address in the actual data (description), in the Megaport one. And I don't see a big problem to lose it by replacing by a generic one.
Maybe I am missing some good point, but so far I think it's easier to do a simple replace to a reference value for all the IP addresses.

itdependsnetworks:
Case in point: 

image

Additionally, I have had bad luck with parsing where a column would be based on 15 characters as an example and then when reduced would run into issues such as:


```
IP               Status
100.100.100.101  active
```

```
IP               Status
192.2.0.1  active
```
and creates a state that would never exist in the real world, but can only exist in the testing infra. So what win? The correct parsing strategy of 15 characters (imagine a real scenario where this is the 4th column and all columns are based on spacing) of the incorrect reality that was created by auto-anonymizing without spacial awareness? 

Long and short, parsing is a fickle beast and I have been personally bitten by this one several times.

glennmatthews:
I'm not as concerned as @itdependsnetworks about specifically preserving exact character counts, but I do agree that it would be good to keep some uniqueness across different IPs rather than making all addresses identical. I'd vote for something like just changing the first octet of IPv4 addresses to `10.` and changing the first two tokens of IPv6 addresses (do we even have any in our examples?) to `2001:db8:`.

anonymize_public_ips.py
```diff
@@ -0,0 +1,111 @@
...
+)
+IPV6_ADDR_REGEX = "|".join([f"(?:{g})" for g in IPV6GROUPS[::-1]])  # Reverse rows for greedy match
+
+
+def replace(filename, search_exp, replace_exp):
+    """Replace line when a substitution is needed."""
+    report = ""
+    try:
+        for line in fileinput.input(filename, inplace=True):
+            if any(pattern_to_skip in line for pattern_to_skip in PATTERNS_TO_SKIP):
```
glennmatthews, line 48: a bit surprising to me that `PATTERNS_TO_SKIP` is a list of exact substring matches, rather than a list of regex patterns.

anonymize_public_ips.py
```diff
@@ -0,0 +1,111 @@
...
+        for file in files:
+            filename = os.path.join(dirpath, file)
+            if not clean:
+                report = ""
+                with open(filename, "r", encoding="utf-8") as test_file:
+                    try:
+                        content = test_file.readlines()
+                    except UnicodeDecodeError as error:
+                        print(f"Warning: Not able to process {filename}: {error}")
+                        continue
```
glennmatthews, line 80: sys.exit here? Or do we have files that we are *expecting* to have decoding errors in?

chadell: yes, the CSV one is not supported, for now. We could solve it at some point if needed

anonymize_public_ips.py
chadell: it could be changed to regex if at some point is needed. For the current need, he simple substring matching was good enough

chadell:
I will give a try to https://github.com/intentionet/netconan
thanks @jvanderaa

scetron (approved):
Looks good to me.

PR #167, opened 2022-09-19
Update GH token

chadell:


chadell:
> This change doesn't seem right to me - why are we using a `NAUTOBOT` token for a non-Nautobot library?

you are right, It looks wrong. Will check if there is need to change to something else or not

chadell:
double checked with the SRE team and `NTC_GITHUB_TOKEN` is the correct one

PR #169, opened 2022-09-24
Anonymize IPs with Netconan

chadell:
This PR uses a different approach than #165 , using the `netconan` library. It makes it easier as less code is needed to maintain, but because it's randomly assigning IPs, it can't be used as a CI check to verify nothing is changing.
Because this is simply a helper script, I think it is still a valid approach, for when anonymization is needed.


tests/unit/data/aws/aws1.eml
```diff
@@ -1,52 +1,52 @@
...
+ Planned Maintenance Notification [AWS Account: 0000000000001]
+MIME-Version: 1.0
+Content-Type: text/plain; charset="UTF-8"
+Content-Transfer-Encoding: quoted-printable
+X-SM-COMMUNICATION: true
+X-SM-COMMUNICATION-TYPE: AWS_DIRECTCONNECT_MAINTENANCE_SCHEDULED
+X-SM-DEDUPING-ID: 7cc8bab7-00bb-44e0-a3ec-bdd1a5560b80-EMAIL--1012261942-036424c1a19ca69ca7ea459ebd6823e1
+Date: Thu, 6 May 2021 21:52:56 +0000
+Feedback-ID: 1.us-east-1.xvKJ2gIiw98/SnInpbS9SQT1XBoAzwrySbDsqgMkBQI=:AmazonSES
+X-SES-Outgoing: 2021.05.06-62.46.238.146
```
scetron, line 11: Is this a randomized IP? This is an actual IP according to WHOIS: https://apps.db.ripe.net/db-web-ui/query?searchtext=62.46.238.146

chadell: The randomisation from `netconan` has no option to define the "usable" IPs to anonymise (or I haven't found it), so it randomly pick "valid" IPs.

glennmatthews (approved):

PR #176, opened 2022-10-21
Handle Zayo table with "Customer Circuit ID" header

jmaslak:
This handles a case where the Zayo table uses "Customer Circuit ID" in the header rather than "Legacy Circuit ID", which caused the parser to crash.


chadell:
Thanks @jmaslak ! 

The testing strategy has to been automated... but today is still manually defined. You should add your test for End to End test, not only the parser. Simply add the test case in https://github.com/networktocode/circuit-maintenance-parser/blob/develop/tests/unit/test_e2e.py#L533

jmaslak:
I have updated the end to end tests now.

chadell (approved):
LGTM

thanks @jmaslak for your first contribution!

PR #177, opened 2022-10-21
Colt parser: use European-style day-first date parsing

jmaslak:
Colt specifies dates with the day first.  I.E. 07/8/2021 is August 7th, not July 8th.


glennmatthews (approved):
Great catch! This should have been obvious from the test emails, as otherwise we're looking at month-long maintenance windows, but clearly we missed it. Thanks for the fix!

PR #179, opened 2022-10-21
Do not require an "[ EXTERNAL ]" marker for Colt email subjects

jmaslak:
The Colt parser was written for systems that add an "[ External ]" indicator to the email subject.  This patch no longer requires that to be there.


circuit_maintenance_parser/parsers/colt.py
```diff
@@ -44,7 +44,7 @@ def parse_subject(self, subject):
         """
         data = {}
         search = re.search(
-            r"\[.+\]\s([A-Za-z\s]+).+?(CRQ\w+-\w+)\s(\d+/\d+/\d+\s\d+:\d+:\d+\s+[A-Z]+).+?(\d+/\d+/\d+\s\d+:\d+:\d+\s+[A-Z]+).+?([A-Z]+)",
+            r"(?:\[.+\]\s)?([A-Za-z\s]+):?\s+?(CRQ\w+-\w+)\s(\d+/\d+/\d+\s\d+:\d+:\d+\s+[A-Z]+).+?(\d+/\d+/\d+\s\d+:\d+:\d+\s+[A-Z]+).+?([A-Z]+)",
```
glennmatthews, line 47: Would you mind adding an example or two of the non `[ EXTERNAL ]` subject lines to the docstrings here and below, just for clarity?

glennmatthews (approved):

glennmatthews (approved):

jmaslak:
I added an example to each -- I don't receive any with the "External" tag, so suspect that was added by some corporate email system for whoever did the initial parser.

chadell (approved):
LGTM

PR #181, opened 2022-10-24
Use latest mypy

chadell:


pyproject.toml
```diff
@@ -45,11 +45,7 @@ yamllint = "^1.20.0"
 bandit = "^1.6.2"
 invoke = "^1.4.1"
 flake8 = "^4.0.0"
-mypy = "^0.961"
-# Dependencies for mypy to correctly analyze code using these libraries
-types-python-dateutil = "^2.8.3"
-types-pytz = "^2022.0.0"
-types-toml = "^0.10.1"
```
glennmatthews, line 52: Looks like these may still be needed? Not sure why you removed them.

glennmatthews (approved):

PR #182, opened 2022-10-25
Add toml to dev dependencies. The new GH setup only install dev dependencies

chadell:
Ref: https://github.com/networktocode/gh-action-setup-poetry-environment/blob/main/action.yml#L8


.github/workflows/ci.yml
```diff
@@ -71,6 +71,8 @@ jobs:
         uses: "actions/checkout@v3"
       - name: "Setup environment"
         uses: "networktocode/gh-action-setup-poetry-environment@v4"
+        with:
+          poetry-install-options: "--with dev"
```
ubajze, line 75: Why not leave the default value?

.github/workflows/ci.yml
```diff
@@ -97,6 +99,8 @@ jobs:
...
```
ubajze, line 103: Same as above

.github/workflows/ci.yml
```diff
@@ -121,6 +125,7 @@ jobs:
         uses: "networktocode/gh-action-setup-poetry-environment@v4"
         with:
           python-version: "${{ matrix.python-version }}"
+          poetry-install-options: "--with dev"
```
ubajze, line 128: Same as above

chadell: I need to install the library to test it, and the default only install the dev, same happened in mypy

poetry.lock
```diff
@@ -805,23 +790,17 @@ optional = false
...
-astroid = [
-    {file = "astroid-2.11.7-py3-none-any.whl", hash = "sha256:86b0a340a512c65abf4368b80252754cda17c02cdbbd3f587dddf98112233e7b"},
-    {file = "astroid-2.11.7.tar.gz", hash = "sha256:bb24615c77f4837c707669d16907331374ae8a964650a66999da3f5ca68dc946"},
-]
-attrs = [
-    {file = "attrs-22.1.0-py2.py3-none-any.whl", hash = "sha256:86efa402f67bf2df34f51a335487cf46b1ec130d02b8d39fd248abfd30da551c"},
-    {file = "attrs-22.1.0.tar.gz", hash = "sha256:29adc2665447e5191d0e7c568fde78b21f9672d344281d0c6e1ab085429b22b6"},
-]
+astroid = []
+attrs = []
```
glennmatthews, line 803: Changes like these (removing a bunch of file hashes from poetry.lock) are usually indicative that you're running an older version of Poetry and/or need to do a `poetry cache clear --all PyPI` command; there was an update to PyPI's API a couple months back that caused Poetry to show this symptom. See https://github.com/python-poetry/poetry/issues/5970 and its related PRs for context.

glennmatthews (approved):

PR #186, opened 2022-10-27
Telia Carrier is now Arelion

jmaslak:


jmaslak:
I left an old class for Telia there, but am glad to remove it -- not sure what the proper way of handing this rename is, but noticed that the library as-is was not processing Arelion maintenance notices so modified it to do that.

README.md
```diff
@@ -58,10 +58,10 @@ By default, there is a `GenericProvider` that support a `SimpleProcessor` using
 
 #### Supported providers using the BCOP standard
 
+- Arelion
```
chadell, line 61:
```suggestion
- Arelion (previously Telia)
```

chadell (approved):
It looks good to me

scetron (dismissed):

chadell (approved):
LGTM

PR #197, opened 2023-01-12
Updating Equinix parser

jasonyates:
* Adding support for additional impact statement
* Adding support for additional notification types. Equinix Connect notifications include 4 columns as IBX name is included. Equinix Fabric notifications include 14 columns as seller and subscriber ID's are included.

### DIFF
--- a/circuit_maintenance_parser/parsers/equinix.py
+++ b/circuit_maintenance_parser/parsers/equinix.py
@@ -104,7 +104,22 @@ class HtmlParserEquinix(Html):
                             account, _, _, circuit = circuit_info  # pylint: disable=unused-variable
                         elif len(circuit_info) == 14:
                             # Equinix Fabric notifications include a lot of additional detail on seller and subscriber ID's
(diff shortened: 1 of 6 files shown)


itdependsnetworks:
helpful hint,  you will need to update the parser to show the case that this is fixing in the tests (to ensure no future regression)

circuit_maintenance_parser/parsers/equinix.py
```diff
@@ -97,7 +99,29 @@ def _parse_table(self, theader_elements, data, impact):  # pylint: disable=no-se
...
+                                _,
+                                _,
+                                _,
+                                _,
+                                _,
+                                _,
+                                _,
+                            ) = circuit_info  # pylint: disable=unused-variable
+                        else:
+                            account, _, circuit = circuit_info  # pylint: disable=unused-variable
```
chadell, line 124: This would protect against exceptions about some circuit_info unexpected structures

```suggestion
                        elif len(circuit_info) == 3:
                            account, _, circuit = circuit_info  # pylint: disable=unused-variable
                        else:
                            return
```

chadell (approved):
LGTM

PR #198, opened 2023-01-12
Fix Verizon parser

jasonyates:
Parser was looking for <tbody> in circuits table. Not all VZ notifications include the <tbody> element.

* Confirmed all existing tests pass with this change.
* Added a new test for a notification without <tbody>

### DIFF
--- a/circuit_maintenance_parser/parsers/verizon.py
+++ b/circuit_maintenance_parser/parsers/verizon.py
@@ -57,7 +57,7 @@ class HtmlParserVerizon1(Html):
                 data["start"] = self.dt2ts(start)
                 data["end"] = self.dt2ts(end)
 
-        for row in circuit_table.find("tbody").find_all("tr"):
+        for row in circuit_table.find_all("tr"):
             cells = row.find_all("td")
             cells_text = [cell.string.strip() for cell in cells if cell.string]
             if not cells_text or cells_text[0].startswith("Company Name"):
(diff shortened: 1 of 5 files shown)


chadell:
If you check other test data result files from Verizon parser, you will see that not all the elements that your result has are extracted.
For instance, https://github.com/networktocode/circuit-maintenance-parser/blob/develop/tests/unit/data/verizon/verizon4_result.json, has no "organization" or "stamp". You should improve the parser to get some of the data, such as the organizer.

These data, for the final notification, is taken from defaults, or via other parsers. For instance, the `stamp` is taken from the date, as you can see in the E2E test:

```
        (
            Verizon,
            [
                ("html", Path(dir_path, "data", "verizon", "verizon3.html")),
                (EMAIL_HEADER_DATE, Path(dir_path, "data", "date", "email_date_1")),
            ],
            [
                Path(dir_path, "data", "verizon", "verizon3_result.json"),
                Path(dir_path, "data", "date", "email_date_1_result.json"),
            ],
        ),
```

So, combining, the HTML parser and a _ficticius_ `email_date_1` (this is simulating the processing of a raw email that takes the date from email headers.

jasonyates:
@chadell Makes sense. Appreciate the additional detail on how the tests run. My additional test now passes correctly locally.

chadell (approved):
LGTM

PR #204, opened 2023-01-27
style: :arrow_up: Update pydantic to 1.10.4

chadell:


jvanderaa (approved):

pyproject.toml
```diff
@@ -24,8 +24,7 @@ include = [
 [tool.poetry.dependencies]
 python = "^3.7"
 click = ">=7.1, <9.0"
-# Skipping 1.9.x versions due issue https://github.com/samuelcolvin/pydantic/issues/4114
-pydantic = {version = "~1.8.2", extras = ["dotenv"]}
+pydantic = {version = ">= 1.8.0, != 1.9.*, >= 1.10.4", extras = ["dotenv"]}
```
glennmatthews, line 27: Remind me, does this actually allow both 1.8.x and 1.10.4+? Or would it be equivalent and simpler to just have `">= 1.10.4"` alone?

glennmatthews (approved):
Hopefully the tests passing means that samuelcolvin/pydantic#4114 isn't an issue for us any more. :)

pyproject.toml
chadell: the idea is to keep the parser working for environments using Pydantic pinned in 1.8.X versions, and start using the newer versions

glennmatthews: Right, just wanted to make sure that's what it's actually doing :)

PR #212, opened 2023-02-23
Updated documentation: Contributing section

csessh:
As I raised the question in this [Slack thread](https://networktocode.slack.com/archives/C0HHE1FFH/p1675731111230349), the `Contributing` section in `README.md` needs updating to make it clearer for those who'd like to quickly debug the library locally. 

I think others have also pointed out that they also got trapped with `invoke build` route which ends up with `no Dockerfile` error message.


chadell (approved):
LGTM

PR #222, opened 2023-03-20
Fix e2e test while combining data from multiple maintenaces 

chadell:


tests/unit/test_e2e.py
```diff
@@ -849,10 +859,18 @@ def test_provider_get_maintenances(
...
+            # TODO: Tests assume that maintenances (multiple) will be discovered on the first parser
+            if not expected_result and isinstance(partial_result_data, list):
                 expected_result = partial_result_data
+
+            if expected_result and isinstance(partial_result_data, dict):
+                for _ in range(len(expected_result)):
+                    expected_result[0].update(partial_result_data)
             else:
-                expected_result[0].update(partial_result_data[0])
+                assert len(expected_result) == len(expected_result)
```
scetron, line 871: Would this not always be true?

chadell: if the expected results are well defined, yes :)

chadell: you're right, typo there

scetron:
This looks good to me!

chadell:
> This looks good to me!

if so, could you approve the PR?

scetron (approved):

PR #227, opened 2023-05-02
Parser for Apple peering notifications

slyngshede:
Parser for Apple peering notification emails. We haven't yet seen anything that can be classified as other than "CONFIRMED"

The EmailSubjectParser behaves slightly different than one would expect based on the naming, in that it just get the entire email text, without line break. I would expect that it would receive ONLY the subject line, but that doesn't seem to be the case. Am I using it wrong perhaps?

### DIFF
--- a/circuit_maintenance_parser/parsers/apple.py
+++ b/circuit_maintenance_parser/parsers/apple.py
@@ -1,0 +1,58 @@
+import email
+import re
+
+from typing import Dict, List
+
+from circuit_maintenance_parser.output import Impact, Status
+from circuit_maintenance_parser.parser import EmailSubjectParser, Text, CircuitImpact
+
+class SubjectParserApple(EmailSubjectParser):
+    def parser_hook(self, raw: bytes):
+        message = email.message_from_string(self.bytes_to_string(raw))
+        return self.parse_subject(message['subject'])
+
+    def parse_subject(self, subject: str) -> List[Dict]:
+        return [{'summary': subject}]
+
+
+class TextParserApple(Text):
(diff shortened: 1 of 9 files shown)


circuit_maintenance_parser/parsers/apple.py
```diff
@@ -0,0 +1,58 @@
...
+            'provider': 'apple',
+        }
+        return [data]
+
+    def _circuits(self, text):
+        pattern = r'Peer AS: (\d*)'
+        m = re.search(pattern, text)
+        return [CircuitImpact(
+            circuit_id=f'AS{m.group(1)}',
+            impact=Impact.NO_IMPACT)]
```
chadell, line 36: I guess you would like to mark the circuit state during the maintenance as OUTAGE? NO_IMPACT is supposed for a circuit that it's not being affected during the maintenance period

chadell:
@slyngshede `black` tests are still complaining
Remember to add the new parser support to the Readme to make it visible.

chadell:
Hi @slyngshede , there is still some linting problem. anything we can help you with?

slyngshede:
@chadell No, I got it working, just been business with other project :-)

chadell:
@slyngshede , if you can't finish it, I could do it for you.

chadell:
@slyngshede , if you can't finish it, I could do it for you.

chadell:
@slyngshede , would you like to complete this? or do you mind if I bring it to the finish line?

slyngshede:
@chadell not at all, please go ahead. I have plenty of other parsers to look at when I find the time again :-)

PR #228, opened 2023-05-25
Remove Python3.7 support.

slyngshede:
Python 3.7 support prevents code from running on newer versions of Python. Update depedencies to work on Python 3.8 and up. 3.7 support makes it impossible to run tests locally on systems with newer versions (3.11 and up) of Python

Remove Cogent from tests, as it's currently broken and depends on an unmaintainced tzwhere library that does not support newer versions of Python.

Disable the one test in test_utils.py as it's only used for Cogent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## v2.2.4 - 2023-07-12
+
+- #230 - Swap out tzwhere for TimezoneFinder
+- #234 - Added upper bound to pydantic dependency
+
+### Changed
+
 ## v2.2.3 - 2023-03-21
 
 ### Changed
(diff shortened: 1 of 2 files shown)


chadell:
Hi @slyngshede , I understand the problem, but we there are users of the library that are still using Python3.7, so we decided to wait until the [End of Life of Python 3.7 in 2023-06-27](2023-06-27) to drop its support.
So, in just one month, we will take this in.
Thanks

chadell:
I see the problem with Cogent, that we will need to tackle

slyngshede:
There is also the option of going through dependencies, and fix them individually. That was me original plan for numpy, but it quickly spirals out of control and it seemed to me that the only major issue was Python 3.7. 

But agreed with the choice to wait until Python 3.7 is officially deprecated, it's soon enough anyway.

chadell:
Hi @slyngshede , would you like to update this PR to get in now that Python3.7 is EOL?

PR #234, opened 2023-07-05
Added upper bound to pydantic dependency

qduk:
Adds an upper bound of `1.X` to pydantic due to current issues with pydantic version 2.X

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ include = [
 [tool.poetry.dependencies]
 python = "^3.7"
 click = ">=7.1, <9.0"
-pydantic = {version = ">= 1.8.0, != 1.9.*, >= 1.10.4", extras = ["dotenv"]}
+pydantic = {version = ">= 1.8.0, != 1.9.*, >= 1.10.4, ^1", extras = ["dotenv"]}
 icalendar = "^5.0.0"
 bs4 = "^0.0.1"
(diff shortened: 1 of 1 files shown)


chadell (dismissed):
LGTM

pyproject.toml
```diff
@@ -24,7 +24,7 @@ include = [
 [tool.poetry.dependencies]
 python = "^3.7"
 click = ">=7.1, <9.0"
-pydantic = {version = ">= 1.8.0, != 1.9.*, >= 1.10.4", extras = ["dotenv"]}
+pydantic = {version = ">= 1.8.0, != 1.9.*, >= 1.10.4, ^1", extras = ["dotenv"]}
```
glennmatthews, line 27: I didn't know you could mix the specifier styles like this actually - would the below be preferable?

```suggestion
pydantic = {version = ">= 1.8.0, != 1.9.*, >= 1.10.4, < 2", extras = ["dotenv"]}
```

glennmatthews (approved):

pyproject.toml
chadell: @qduk I also prefer @glennmatthews suggestion. could you check?

qduk: That will work. Updated

glennmatthews (approved):

PR #243, opened 2023-10-20
Handle broken EUNetworks cancellation messages

jmaslak:
EUNetworks sends at least some cancellation messages without any circuits included in the message.  I've included a redacted sample email demonstrating this.  Other EUNetworks message types seem to have the proper structure.  To solve this, I inserted a bogus circuit into the parsed output (only when the status type is a cancel status), but I'm open to other suggestions if it would be better if I did it a different way.


circuit_maintenance_parser/parser.py
```diff
@@ -138,13 +138,20 @@ def parse_ical(gcal: Calendar) -> List[Dict]:
...
-                else:
+                elif circuits is not None:
                     data["circuits"] = [
                         CircuitImpact(
                             circuit_id=circuits,
                             impact=Impact(component.get("X-MAINTNOTE-IMPACT")),
                         )
                     ]
+                elif data["status"] == "CANCELLED":
+                    # We shouldn't have to do this, as it does not
```
chadell, line 149: trying to understand...
EUNetworks is sending a notification with a maintenance ID about canceling it, but it's not providing any information about the circuit IDs, which were referenced in the initial notification message. isn't it?

jmaslak: Yes, you understand properly.  The actual maintenance notification draft document however says that most information should be copied from the previous notification when sending a cancelled status - https://datatracker.ietf.org/doc/html/draft-gunter-calext-maintenance-notifications-00#section-3.3 - so hopefully this is rare behavior for at least the people following that draft (I've only seen it from EUNetworks but that of course doesn't mean nobody else is doing it).

I'll submit an update to this PR that does this as you describe.

jmaslak:
I reworked this PR based on the comments above.  I also updated some incorrect information in the comments (namely that status is a required field, not optional).  I did have to move the status field around in output.py for the empty circuit list validator to be able to access it (it had to be specified before circuits was).

chadell:
> I reworked this PR based on the comments above. I also updated some incorrect information in the comments (namely that status is a required field, not optional). I did have to move the status field around in output.py for the empty circuit list validator to be able to access it (it had to be specified before circuits was).

Thanks @jmaslak for your contribution!

PR #244, opened 2023-10-20
AWS parser update

garretts-ntc:
First PR for feedback and assistance with building/designing tests! 🙏


circuit_maintenance_parser/parsers/aws.py
```diff
@@ -23,10 +24,25 @@ def parse_subject(self, subject):
...
+        regex_keys = re.compile("|".join(subject_map), re.IGNORECASE)
+
+        # in case of a multi-line subject
+        # match the subject map
+        for line in subject.splitlines():
+            line_matched = re.search(regex_keys, line)
+            if not line_matched:
+                continue
+            for group_match in line_matched.groups():
+                if group_match is not None:
```
itdependsnetworks, line 42: what about

```python
if not group_match:
    continue
```

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -13,6 +13,7 @@
 # pylint: disable=too-many-nested-blocks, too-many-branches
 
 logger = logging.getLogger(__name__)
+#logger.setLevel("DEBUG")
```
itdependsnetworks, line 16: can remove.

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -60,29 +76,63 @@ def parse_text(self, text):
...
+
+                if re.match(r"[a-z]{5}-[a-z0-9]{8}", line):
+                    maintenace_id += line
+                    data["circuits"].append(CircuitImpact(circuit_id=line, impact=impact))
+                continue
+
+            # for lines that do match our regex strings.
+            # grab the data and map the values to keys.
+            for group_match in line_matched.groups():
+                if group_match is not None:
```
itdependsnetworks, line 115: similar to above.

garretts-ntc: nice! yes this let me shore up one indentation. I'm definitely needing to explore guardian clauses.

circuit_maintenance_parser/parsers/aws.py
garretts-ntc: noice! :D

circuit_maintenance_parser/parsers/aws.py
garretts-ntc: see below ;D - guardian clause implemented.

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -23,10 +23,26 @@ def parse_subject(self, subject):
 
         Example: AWS Direct Connect Planned Maintenance Notification [AWS Account: 00000001]
         """
-        data = {}
-        search = re.search(r"\[AWS Account ?I?D?: ([0-9]+)\]", subject)
-        if search:
-            data["account"] = search.group(1)
+        data = {"account": ""}
```
chadell, line 26: If we can't find the "account", we should skip adding the key with an empty value. It's better to fail clearly than have a result like the one in `tests/unit/data/aws/aws3_subject_parser_result.json`

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -23,10 +23,26 @@ def parse_subject(self, subject):
...
+        for line in subject.splitlines():
+            line_matched = re.search(regex_keys, line)
+            if not line_matched:
+                continue
+            for group_match in line_matched.groups():
+                if not group_match:
+                    continue
+                for key, value in subject_map.items():
+                    if re.search(key, line, re.IGNORECASE):
+                        data[value] = group_match
```
chadell, line 45: I know it works and could cover more future REGEX matching besides "account", but I guess it was not necessary to introduce a map for this unique key, no? (no need to change)

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -60,31 +76,67 @@ def parse_text(self, text):
...
+        text_map = {
+            "^Account ?I?D?: ([0-9]+)": "account",
+            "^Start Time: ([A-Z][a-z]{2}, [0-9]{1,2} [A-Z][a-z]{2,9} [0-9]{4} [0-9]{2}:[0-9]{2}:[0-9]{2} [A-Z]{2,3})": "start",
+            "^End Time: ([A-Z][a-z]{2}, [0-9]{1,2} [A-Z][a-z]{2,9} [0-9]{4} [0-9]{2}:[0-9]{2}:[0-9]{2} [A-Z]{2,3})": "end",
+            "(?<=from )([A-Z][a-z]{2}, [0-9]{1,2} [A-Z][a-z]{2,9} [0-9]{4} [0-9]{2}:[0-9]{2}:[0-9]{2} [A-Z]{2,3}) to ([A-Z][a-z]{2}, [0-9]{1,2} [A-Z][a-z]{2,9} [0-9]{4} [0-9]{2}:[0-9]{2}:[0-9]{2} [A-Z]{2,3})": "start_and_end",
+        }
+
+        regex_keys = re.compile("|".join(text_map), re.IGNORECASE)
+
+        data = {"circuits": [], "start": "", "end": ""}
```
chadell, line 88: again, I prefer to not start with empty "start" and "end" values, if we can't solve them, let's leave them empty

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -60,31 +76,67 @@ def parse_text(self, text):
...
+            for group_match in line_matched.groups():
+                if not group_match:
+                    continue
+                for key, value in text_map.items():
+                    if re.search(key, line_matched.string, re.IGNORECASE):
+                        # Due to having a single line on some emails
+                        # This causes multiple match groups
+                        # However this needs to be split across keys.
+                        # This could probably be cleaned up.
+                        if value == "start_and_end" and data["start"] == "":
```
chadell, line 118: this assumes that start date is always first, isn't it?
you could do similar check without the empty initialization with:
```suggestion
                        if value == "start_and_end" and "start" not in data:
```

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -60,31 +76,67 @@ def parse_text(self, text):
...
+                            data[value] = group_match
+
+            # Let's determine impact and status
+            if "may become unavailable" in line.lower():
+                impact = Impact.OUTAGE
+            elif "has been cancelled" in line.lower():
+                status = Status.CANCELLED
+
+        # Let's get our times in order.
+        if data["start"] and data["end"]:
```
chadell, line 132:
```suggestion
        if all([key in data.keys() for key in ["start", "end"]]):
```

circuit_maintenance_parser/parsers/aws.py
garretts-ntc: I guess that's the rub in this case. Some of the emails have the account number in the subject line, while others don't, (it's in the body).  If I just skip adding an empty key, there is no other data thats parsed on subject line, and it returns an empty data structure `[{}]` this breaks the upstream parser tho, as it considers it to have failed parsing vs. an expected empty response.  Maybe I'm not understanding it clearly tho' - what would be a good practice to do here when you are EXPECTING there to be no data under certain conditions?

circuit_maintenance_parser/parsers/aws.py
garretts-ntc: Yea, when I was extending the text parser, i realized I could do the same thing on subject, and thought that it might make sense just to do the same on both parsers, that way, in case this case ever happened again, it would just be a quick regex pattern match. ... But then after doing it, I realized that there was only the match.... but I by that time I was committed, plus it felt like sense to keep the same pattern in the provider.

circuit_maintenance_parser/parsers/aws.py
garretts-ntc: Excellent. This is much better.

circuit_maintenance_parser/parsers/aws.py
garretts-ntc: Incorporated.

circuit_maintenance_parser/parsers/aws.py
chadell: the `Providers` (e.g., aws), contain different `Processors`, which are a combination of one or more `Parsers`.
The way the circuit maintenance parser works is to try to extract part of the `Maintenance` data from each `Parser` and then finally combine the data to create a valid `Maintenance`.
This means that is some `Processor` has a `Parser` that is not capturing the proper data it's ok, another one could do in the same process.

gskjelstad: Alright cool, I think after incorporating the feedback you gave me, I think we probably are ready to give this a look over again. Let me know what you think! 👍

circuit_maintenance_parser/provider.py
```diff
@@ -186,6 +186,7 @@ class AWS(GenericProvider):
     """AWS provider custom class."""
 
     _processors: List[GenericProcessor] = [
+        CombinedProcessor(data_parsers=[EmailDateParser, TextParserAWS1]),
```
chadell, line 189: have you tried to invert them? so we test first the most complete parsing, including the Subject, and only then it fails, we try the second approach.

gskjelstad: yep! I think we had even hinted in that direction, to ensure that the more complete parser would take precedent in it's processing.  Thanks for the reminder! I've swapped those!

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -60,31 +80,79 @@ def parse_text(self, text):
...
-            elif re.match(r"[a-z]{5}-[a-z0-9]{8}", line):
-                maintenace_id += line
-                data["circuits"].append(CircuitImpact(circuit_id=line, impact=impact))
+            # match against the regex strings
+            line_matched = re.search(regex_keys, line)
+            # if we have a string that's not in our text_map
+            # there may still be some strings with data to capture.
+            # otherwise, continue on.
+            if not line_matched:
+                if re.match(r"[a-z]{5}-[a-z0-9]{8}", line):
```
chadell, line 117: just curious why haven't you added this Regex to the others?

circuit_maintenance_parser/parsers/aws.py
```diff
@@ -24,9 +24,29 @@ def parse_subject(self, subject):
         Example: AWS Direct Connect Planned Maintenance Notification [AWS Account: 00000001]
         """
         data = {}
-        search = re.search(r"\[AWS Account ?I?D?: ([0-9]+)\]", subject)
-        if search:
-            data["account"] = search.group(1)
+        # Common Subject strings for matching:
+        subject_map = [{"account": r"\[AWS Account ?I?D?: ([0-9]+)\]"}]
+
+        subject_list = []
```
chadell, line 30: this is the exactly same code you have in the other function, extract this to a staticmethod to reuse

Issue #225, opened 2023-03-28
Investigate using LLM (e.g. ChapGPT) as a `Processor`


chadell:
### Proposed Functionality

New LLM (Large Language Model) allow a sophisticated language processing that could get data from structures/unstructured notifications and retrieve part or total of the necessary information.

### Use Case

This could avoid having to write custom parser for all Providers

chadell:
Yes, my idea would be to make it the last resort when the known parser failed or there is no parser at all.
Preliminary testing looks good!

KedharnathGoud:
We can use something like https://github.com/gventuri/pandas-ai

chadell:
Hi @AndriusV4 , this is something we definitely want to explore, but we have not started yet. 
Any contribution in this direction would be welcome!

chadell:
I already have PoC that allows parsing only using Machine Learning:

```json
$ circuit-maintenance-parser --data-file "tests/unit/data/aws/aws1.eml" --data-type email 

Circuit Maintenance Notification #0
{
  "account": "aaaaa",
  "circuits": [
    {
      "circuit_id": "aaaaa-00000001",
      "impact": "OUTAGE"
    },
    {
      "circuit_id": "aaaaa-00000002",
      "impact": "OUTAGE"
    },
    {
      "circuit_id": "aaaaa-00000003",
      "impact": "OUTAGE"
    },
    {
      "circuit_id": "aaaaa-00000004",
      "impact": "OUTAGE"
    },
    {
      "circuit_id": "aaaaa-00000005",
      "impact": "OUTAGE"
    },
    {
      "circuit_id": "aaaaa-00000006",
      "impact": "OUTAGE"
    }
  ],
  "end": 1621522800,
  "maintenance_id": "d9f256dcae6c34a4d7c353b19f02957c",
  "organizer": "unknown",
  "provider": "genericprovider",
  "sequence": 1,
  "stamp": 1620337976,
  "start": 1621497600,
  "status": "CONFIRMED",
  "summary": "Planned maintenance has been scheduled on an AWS Direct Connect router in A Block, New York, NY from Thu, 20 May 2021 08:00:00 GMT to Thu, 20 May 2021 14:00:00 GMT for 6 hours.",
  "uid": "0"
}

chadell:
done in #245

PR #245, opened 2023-10-20
feat: :sparkles: OpenAI parser

chadell:
This PR tries to answer #225 , establishing a pattern to use LLM APIs (such as OpenAI) as last resource parsers.

TODO:

- [x] Documentation: how to use it (ENV), expected behavior, warnings
- [x] Tests: mock response
- [x] New Processor: when HTML and text available, only one used to not duplicate


chadell:
looking forward to your feedback!

README.md
```diff
@@ -86,6 +86,20 @@ By default, there is a `GenericProvider` that support a `SimpleProcessor` using
 
 > Note: Because these providers do not support the BCOP standard natively, maybe there are some gaps on the implemented parser that will be refined with new test cases. We encourage you to report related **issues**!
 
+#### LLM-powered Parsers
+
+The library supports an optional parser option leveraging Large Language Model (LLM) to provide a best-effort parsing when the specific parsers have not been successful.
+
+These LLM parsers are automatically appended as a processor option after the already available for all the Providers when the integration environmental variable is set (check the below integrations).
```
glennmatthews, line 93: Having a bit of trouble parsing this. perhaps?

```suggestion
When the appropriate environment variable(s) are set (see below), these LLM parsers are automatically appended after all existing processors for each defined Provider.
```

circuit_maintenance_parser/parser.py
```diff
@@ -261,3 +263,144 @@ def get_text_hook(raw: bytes) -> str:
...
+
+    _data_types = ["text/html", "html", "text/plain"]
+
+    _llm_question = (
+        "Can you extract the maintenance_id, the account_id, the impact, the status "
+        "(e.g., confirmed, cancelled, rescheduled), the summary, the circuit ids (also defined as service or order), "
+        "and the global start_time and end_time as EPOCH timestamps in JSON format with keys in "
+        "lowercase underscore format? Reply with only the answer in JSON form and "
+        "include no other commentary"
+    )
```
glennmatthews, line 279: Should this be moved to the OpenAPI parser specifically or at least made overridable? I could see that different LLMs might need variant phrasings of the question to get the desired outcome.

circuit_maintenance_parser/parser.py
```diff
@@ -261,3 +263,144 @@ def get_text_hook(raw: bytes) -> str:
...
+            content = soup.text
+        else:
+            content = self.get_text_hook(raw)
+        for data in self.parse_content(content):
+            result.append(data)
+        return result
+
+    @staticmethod
+    def get_text_hook(raw: bytes) -> str:
+        """Can be overwritten by subclasses."""
```
glennmatthews, line 295: What purpose does this method serve? When would subclasses need/want to override it?

circuit_maintenance_parser/parser.py
```diff
@@ -261,3 +263,144 @@ def get_text_hook(raw: bytes) -> str:
...
+        return account
+
+    def _get_maintenance_id(self, generated_json: dict, start, end, circuits):
+        """Method to get the Maintenance ID."""
+        maintenance_key = self.get_key_with_string(generated_json, "maintenance")
+        if maintenance_key and generated_json["maintenance_id"] != "N/A":
+            return generated_json["maintenance_id"]
+
+        maintenace_id = str(start) + str(end) + "".join(list(circuits))
+        return hashlib.md5(maintenace_id.encode("utf-8")).hexdigest()  # nosec
```
glennmatthews, line 378:
```suggestion
        maintenance_id = str(start) + str(end) + "".join(list(circuits))
        return hashlib.md5(maintenance_id.encode("utf-8")).hexdigest()  # nosec
```

tests/unit/test_parsers.py
```diff
@@ -544,7 +544,9 @@ def test_parsers(parser_class, raw_file, results_file):
     with open(raw_file, "rb") as file_obj:
         raw_data = file_obj.read()
 
-    parsed_notifications = parser_class().parse(raw_data)
+    parsed_notifications = parser_class().parse(
+        raw_data, parser_class._data_types[0]  # pylint: disable=protected-access
+    )
```
glennmatthews, line 549: Can you help me understand why this change? It feels weird to be passing a private attribute of the `parser_class` into a method on said class - why doesn't the parser just look this up automatically?

circuit_maintenance_parser/parser.py
chadell: My idea is to have a generic one that is used by default, and then, every LLM could overwrite as needed

circuit_maintenance_parser/processor.py
```diff
@@ -54,13 +54,13 @@ def process(self, data: NotificationData, extended_data: Dict) -> Iterable[Maint
         self.extended_data = extended_data
         maintenances_data: List = []
 
-        # First, we generate a list of tuples with a `DataPart` and `Parser` if the data type from the first is
-        # supported by the second.
-        data_part_and_parser_combinations = [
-            (data_part, data_parser)
+        # First, we generate a set with the key `Parser` and `DataPart` if the data type from the first is
```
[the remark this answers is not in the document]
chadell: yes

circuit_maintenance_parser/provider.py
```diff
@@ -122,6 +123,9 @@ def get_maintenances(self, data: NotificationData) -> Iterable[Maintenance]:
             logger.debug("Skipping notification %s due filtering policy for %s.", data, self.__class__.__name__)
             return []
 
+        if os.getenv("OPENAI_TOKEN"):
+            self._processors.append(CombinedProcessor(data_parsers=[EmailDateParser, OpenAIParser]))
```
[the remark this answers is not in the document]
chadell: because the "Stamp" (when the notification was received) it's not available in the notification body. This parser allows extracting the date from the notification without the need of ML processing.

tasks.py
```diff
@@ -262,13 +262,20 @@ def tests(context, local=INVOKE_LOCAL):
         context (obj): Used to run specific commands
         local (bool): Define as `True` to execute locally
     """
+    token_openai = os.getenv("OPENAI_TOKEN")
+    if token_openai:
+        del os.environ["OPENAI_TOKEN"]
```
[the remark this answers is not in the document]
chadell: yes, that's for my local testing. I will consider adding a conftest to mock it.

tests/unit/test_parsers.py
chadell: before this PR, the Parsers were only able to manage one data type: HTML, CSV, text, etc. The LLM Parser supports two types: HTML and text, but it requires a specific transform of the data into text (done [here](https://github.com/networktocode/circuit-maintenance-parser/pull/245/files#diff-b5df3420c13a1de156bbb25c6f41d0448704d43bd9feee7f0bcfe131804f0f8eR281)). 

However, I should change this private access to the public method to get them.

circuit_maintenance_parser/parser.py
chadell: I made it more explicit why it is used. the final parsers could have a different way to decode bytes to string

jvanderaa (dismissed):

pyproject.toml
```diff
@@ -32,6 +32,7 @@ geopy = "^2.1.0"
 timezonefinder = "^6.0.1"
 backoff = "^1.11.1"
 chardet = "^5"
+openai = "^0.28.1"
```
glennmatthews, line 35: How heavyweight is this dependency? Would it be preferable to make it an optional extra instead of a hard requirement?

glennmatthews (approved):

pyproject.toml
chadell: It doesn't look crazy: https://github.com/openai/openai-python/blob/main/pyproject.toml#L10
However, I think that making it optional is an extra safeguard to make people aware of using it
And, I also noticed that there is a stable version!

glennmatthews (approved):

Issue #246, opened 2023-10-21
Incorporate Metadata to the Maintenance response


chadell:
### Environment

- circuit_maintenance_parser version: 

### Proposed Functionality

Besides the data representing the Maintenance, the response could incorporate metadata to understand things like:
- Which Provider has been used to parse it?
- Which Parsers have been used?
- If the parsing consumed some resources, a summary of the amount of them (for instance, Openai tokens)

chadell:
working on #249

chadell:
Completed by #249

PR #249, opened 2023-11-15
Add Metadata to every Maintenance, addresses issue #246

chadell:
The idea is to provide some basic information about how the Maintenance has been generated to allow consumers to take decisions

## TODO

- [x] Update Readme

Example of the information retrieved via CLI:

image


glennmatthews (commented):
Looks like a reasonable enhancement to me!

chadell:
> Looks like a reasonable enhancement to me!

nice, I have updated the readme to explain it.

circuit_maintenance_parser/parser.py
```diff
@@ -43,6 +43,11 @@ def get_data_types(cls) -> List[str]:
         """Return the expected data type."""
         return cls._data_types
 
+    @classmethod
+    def get_name(cls) -> str:
+        """Return the parse name."""
```
glennmatthews, line 48:
```suggestion
        """Return the parser name."""
```

glennmatthews (dismissed):

circuit_maintenance_parser/output.py
```diff
@@ -139,12 +154,18 @@ class Maintenance(BaseModel, extra=Extra.forbid):
...
+    _metadata: Metadata = PrivateAttr()
 
     # Non mandatory attributes
     uid: StrictStr = "0"
     sequence: StrictInt = 1
     summary: StrictStr = ""
 
+    def __init__(self, **data):
+        """Initialize the Maintenance object."""
+        self._metadata = data.pop("_metadata")
```
csessh, line 166: This throws KeyError if _metadata key isn't present, thus this change becomes a breaking change. 

Some of my existing tests used in my projects start to throw KeyError exception. For example:

```
self = Maintenance()
data = {'account': '12623', 'circuits': [{'circuit_id': '297487', 'impact': 'NO-IMPACT'}], 'end': 1649073600, 'maintenance_id': 'GIN-CHG0042605', ...}

    def __init__(self, **data):
        """Initialize the Maintenance object."""
>       self._metadata = data.pop("_metadata")
E       KeyError: '_metadata'

../../dev/MNA/lib/python3.10/site-packages/circuit_maintenance_parser/output.py:166: KeyError
```

csessh: It seems like Maintenance dataclass is now expecting _metadata to be supplied by users. can't it default to None?

chadell: thanks @csessh for reporting. I'm going to check and fix.

chadell: checking the code, I see that the `_metadata` is injected by the `Processors` automatically, so I don't get why don't you get it in your tests.

PR #251, opened 2023-12-08
Add missing Arelion entry to SUPPORTED_PROVIDERS

glennmatthews:
Longer term it would be nice to replace this omission-prone hard-coded list and instead do something like automatically pulling in all non-abstract subclasses of `GenericProvider`.


chadell (approved):
LGTM

PR #252, opened 2023-12-12
feat: :sparkles: Allow customization of LLM question

chadell:


circuit_maintenance_parser/parser.py
```diff
@@ -346,6 +347,14 @@ def get_key_with_string(dictionary: dict, string: str):
                 return key
         return None
 
+    def get_llm_question(self):
+        """Return the LLM question."""
+        custom_llm_question = os.getenv("OPENAI_API_KEY")
```
glennmatthews, line 352: Doesn't look like the right variable to me :)

Is an environment variable the appropriate way to do this, given that the question can be a very long string with whitespace and newlines?

chadell: well, maybe it's not the most user-friendly one, but the library has no config options so far... do you think it deserves another thought?

chadell, line 352:
```suggestion
        custom_llm_question = os.getenv("LLM_QUESTION")
```

glennmatthews: I wonder if something like a path to a text file might be preferable. Not sure though.

chadell: this could be an option. Actually, we could offer both (in precedence order):

1. `LLM_QUESTION_STR`
2. `LLM_QUESTION_FILEPATH`

chadell: @glennmatthews I implemented this approach.

README.md
```diff
@@ -98,9 +98,12 @@ When the appropriate environment variable(s) are set (see below), these LLM pars
...
 These are the currently supported LLM integrations:
 
+- `PARSER_LLM_QUESTION_STR` (Optional), question to overwrite the default one. Change it carefully. It has precedence over `PARSER_LLM_QUESTION_FILEPATH`
+- `PARSER_LLM_QUESTION_FILEPATH` (Optional), a path to a file that contains a question to overwrite the default one.
+
 - [OpenAI](https://openai.com/product), these are the supported ENVs:
-  - `OPENAI_API_KEY` (Required): OpenAI API Key.
-  - `OPENAI_MODEL` (Optional): The LLM model to use, defaults to "gpt-3.5-turbo".
+  - `PARSER_OPENAI_API_KEY` (Required): OpenAI API Key.
+  - `PARSER_OPENAI_MODEL` (Optional): The LLM model to use, defaults to "gpt-3.5-turbo".
```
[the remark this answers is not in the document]
chadell: not shipped yet

circuit_maintenance_parser/parsers/openai.py
```diff
@@ -24,15 +24,15 @@ def get_llm_response(self, content) -> Optional[List]:
...
+        client = OpenAI(api_key=os.getenv("PARSER_OPENAI_API_KEY"))
+        model = os.getenv("PARSER_OPENAI_MODEL", "gpt-3.5-turbo")
         try:
             response = client.chat.completions.create(
                 model=model,
                 messages=[
                     {  # type: ignore
                         "role": "system",
-                        "content": self._llm_question,
+                        "content": self.get_llm_question,
```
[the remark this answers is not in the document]
chadell: you were right, in some iteration I forgot to save the `@property`. Tested and working fine now

glennmatthews (approved):

circuit_maintenance_parser/parser.py
```diff
@@ -349,10 +349,18 @@ def get_key_with_string(dictionary: dict, string: str):
 
     def get_llm_question(self):
         """Return the LLM question."""
-        custom_llm_question = os.getenv("LLM_QUESTION")
+        custom_llm_question = os.getenv("PARSER_LLM_QUESTION_STR")
```
jvanderaa, line 353: What if we have multiple parsers? Like I have subscriptions to WatsonX, ChatGPT, and Bard?

tests/conftest.py
```diff
@@ -2,16 +2,16 @@
 import os
 
 
-token_openai = os.getenv("OPENAI_API_KEY")
+token_openai = os.getenv("PARSER_OPENAI_API_KEY")
```
jvanderaa, line 5: I think we could do some refactoring to take the environment key out of testing.

jvanderaa (commented):
I'll take a deeper look to see if I can help with the testing portions.

circuit_maintenance_parser/parser.py
chadell: The LLM question is the same for all

tests/conftest.py
chadell: this was to allow you to have it in your dev environment but not use it on the tests. any better idea?

chadell: I'm gonna merge this PR as this is not a related topic, but please create an issue to improve it later

PR #256, opened 2023-12-21
Improve Equinix parsing

jmaslak:
This makes five changes:

1. It now handles cancel messages.
2. Equinix sometimes uses `<strong>` instead of `<bold>`, so I updated the beautifulsoup calls to look for both (and changed` _parse_b()` to `_parse_bolded()` to be more accurate).
3. UTC: was never found in my sample messages in `b_elem`, but it was always in `b_elem.text`, so I updated that. Tests still pass and it parses the messages I am receiving from Equinix.
4. "There will be service interruptions" was found in `impact_line` in my messages, not i`mpact_line.next_sibling` (which was None and thus causing an exception).  I reworked it to do both checks, and also not crash if `impact_line` does not have a `next_sibling`.
5. I added tests for the new code, using messages I've received from Equinix as the source, with obfuscation applied.

### DIFF
--- a/circuit_maintenance_parser/parsers/equinix.py
+++ b/circuit_maintenance_parser/parsers/equinix.py
@@ -23,7 +23,8 @@ class HtmlParserEquinix(Html):
         """
         data: Dict[str, Any] = {"circuits": []}
 
-        impact = self._parse_b(soup.find_all("b"), data)
+        bolded_elems = soup.find_all(["b", "strong"])
+        impact = self._parse_bolded(bolded_elems, data)
         self._parse_table(soup.find_all("th"), data, impact)
         return [data]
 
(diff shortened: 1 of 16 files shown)


jmaslak:
Sorry about the test failures - I've fixed them.

jmaslak:
I adjusted the next_sibling check for impacts based on a new Equinix message that just came through.  I added a test case for this one as well.

jmaslak:
Last change (really) is to allow multi-language messages where the first part is in a language that looks like ASCII.  I did this by verifying the year is set before processing the maintenance times.  If it's not set, we are likely processing the wrong language and should wait for the right one.  I did add test cases for this as well.

chadell (approved):
LGTM
thank you very much @jmaslak for your contributions!

PR #259, opened 2023-12-27
Feature/crown castle fiber

jmaslak:
Note that unfortunately Crown Castle sometimes sends completion notices w/o circuit IDs.  This is similar to EUNetworks sending cancellation notices without circuit IDs, and I resolved this the same way -- see PR #243.


chadell (approved):
LGTM, thank you!

aliex-13:
Hi @chadell / @jmaslak ,

Just looking into this MR - Was there a typo when you added in "COMPLETED" to the  'validate_empty_circuits' function?

I can see in feature #243, we had the following change:

```
        if len(value) < 1 and values["status"] != "CANCELLED":
            raise ValueError("At least one circuit has to be included in the maintenance")
```


Meaning if the length of the circuit list was 0 (no entries), and the status was not a cancellation, THEN raise a ValueError - which was the correct logic.


After this MR (#259) I can see that code has changed:

```
        if len(value) < 1 and str(values["status"]) in ("CANCELLED", "COMPLETED"):
            raise ValueError("At least one circuit has to be included in the maintenance")
```

Meaning if the length of the circuit list was 0 (no entries), and the status WAS in 'Cancelled' or 'Completed', THEN raise a ValueError.... Which I believe is incorrect. This should be:

```
        if len(value) < 1 and str(values["status"]) not in ("CANCELLED", "COMPLETED"):
            raise ValueError("At least one circuit has to be included in the maintenance")
```


Please correct me if I am wrong but I just noticed this for some of our other vendors.

jmaslak:
I think you are right @aliex-13 - I'll take a look at this tomorrow.

PR #272, opened 2024-02-28
Correct incorrect logic in validation code

jmaslak:
This fixes the issue reported by https://github.com/networktocode/circuit-maintenance-parser/pull/259#issuecomment-1965547758

While the test code validated that this was working, the bad logic worked because `str(values["status"])` would be prefaced by `Status.`.  So I corrected both the list and the logical operator.

### DIFF
--- a/circuit_maintenance_parser/output.py
+++ b/circuit_maintenance_parser/output.py
@@ -197,7 +197,7 @@ class Maintenance(BaseModel, extra="forbid"):
     def validate_empty_circuits(cls, value, values):
         """Validate non-cancel notifications have a populated circuit list."""
         values = values.data
-        if len(value) < 1 and str(values["status"]) in ("CANCELLED", "COMPLETED"):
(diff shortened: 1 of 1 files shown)


circuit_maintenance_parser/output.py
```diff
@@ -197,7 +197,7 @@ def validate_empty_strings(cls, value):
     def validate_empty_circuits(cls, value, values):
         """Validate non-cancel notifications have a populated circuit list."""
         values = values.data
-        if len(value) < 1 and str(values["status"]) in ("CANCELLED", "COMPLETED"):
+        if len(value) < 1 and str(values["status"]) not in ("Status.CANCELLED", "Status.COMPLETED"):
```
chadell, line 200: what if you just compare the enum value directly?
```suggestion
        if len(value) < 1 and values["status"] not in (Status.CANCELLED, Status.COMPLETE):
```

chadell (approved):
LGTM

PR #280, opened 2024-03-26
Add new Windstream Parser

aliex-13:
Raising this Pull Request to onboard a new vendor - Windstream.

All testing I have completed looks to deal with these emails correctly.

Invoke tests --local does report that I have too many local variables but I am not sure how strict we are being on these things as I am seeing messages for other parsers in there as well.

### DIFF
--- a/circuit_maintenance_parser/parsers/windstream.py
+++ b/circuit_maintenance_parser/parsers/windstream.py
@@ -13,7 +13,6 @@ class HtmlParserWindstream1(Html):
     """Notifications Parser for Windstream notifications."""
 
     def parse_html(self, soup):
-
         """Execute parsing."""
         data = {}
         data["circuits"] = []
--- a/circuit_maintenance_parser/parsers/windstream.py
+++ b/circuit_maintenance_parser/parsers/windstream.py
@@ -1,0 +1,78 @@
+"""Windstream Fiber parser."""
+import logging
+from datetime import datetime, timezone
+
+from circuit_maintenance_parser.parser import Html, Impact, CircuitImpact, Status
+
+# pylint: disable=too-many-nested-blocks, too-many-branches
+
+logger = logging.getLogger(__name__)
+
+
+class HtmlParserWindstream1(Html):
+    """Notifications Parser for Windstream notifications."""
+
+    def parse_html(self, soup):
+
+        """Execute parsing."""
+        data = {}
+        data["circuits"] = []
+        impact = Impact("NO-IMPACT")
+        confirmation_words = [
+            "Demand Maintenance Notification",
+            "Planned Maintenance Notification",
+            "Emergency Maintenance Notification",
+        ]
+        cancellation_words = ["Postponed Maintenance Notification", "Cancelled Maintenance Notification"]
+
+        h1_tag = soup.find("h1")
+        if h1_tag.string.strip() == "Completed Maintenance Notification":
+            data["status"] = Status("COMPLETED")
+        elif any(keyword in h1_tag.string.strip() for keyword in confirmation_words):
+            data["status"] = Status("CONFIRMED")
+        elif h1_tag.string.strip() == "Updated Maintenance Notification":
+            data["status"] = Status("RE-SCHEDULED")
+        elif any(keyword in h1_tag.string.strip() for keyword in cancellation_words):
+            data["status"] = Status("CANCELLED")
+
+        div_tag = h1_tag.find_next_sibling("div")
+        summary_text = div_tag.get_text(separator="\n", strip=True)
+        summary_text = summary_text.split("\nDESCRIPTION OF MAINTENANCE")[0]
+
+        data["summary"] = summary_text
+
+        table = soup.find("table")
+        for row in table.find_all("tr"):
+            if len(row) < 2:
+                continue
+            cols = row.find_all("td")
+            header_tag = cols[0].string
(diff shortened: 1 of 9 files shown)


circuit_maintenance_parser/parsers/windstream.py
```diff
@@ -0,0 +1,77 @@
...
+            if header_tag == "WMT:":
+                data["maintenance_id"] = value_tag
+            elif header_tag == "Event Start Date & Time:":
+                value_tag = value_tag.replace(" ET", "")
+                value_tag = datetime.strptime(value_tag, "%m/%d/%y %H:%M")
+                data["start"] = int(value_tag.replace(tzinfo=timezone.utc).timestamp())
+            elif header_tag == "Event End Date & Time:":
+                value_tag = value_tag.replace(" ET", "")
+                value_tag = datetime.strptime(value_tag, "%m/%d/%y %H:%M")
+                data["end"] = int(value_tag.replace(tzinfo=timezone.utc).timestamp())
```
chadell, line 62: what about?
```suggestion
            elif "Date & Time:" in header_tag:
                value_tag = value_tag.replace(" ET", "")
                value_tag = datetime.strptime(value_tag, "%m/%d/%y %H:%M")
                event_time =  int(value_tag.replace(tzinfo=timezone.utc).timestamp())
                if "Event Start" in header_tag:
                    data["start"] = event_time
                elif "Event End" in header_tag:
                    data["end"] = event_time
```

chadell (dismissed):
LGTM

chadell:
> Invoke tests --local does report that I have too many local variables but I am not sure how strict we are being on these things as I am seeing messages for other parsers in there as well.

Don't worry about it

circuit_maintenance_parser/parsers/windstream.py
aliex-13: Yeah that should work! I'll add it now and test it out.

Note: Tested and works fine :)

aliex-13:
Hey @chadell - I just noticed that my start/end logic was incorrect as it was marking the initial time received from the vendor as UTC from the get-go. I have modified the code to correctly localise it to "EST". Conversions to UTC will then occur correctly. This did mean that I needed to move some things around as I added a bunch of extra local variables to deal with it. I have managed to avoid going over the number of local variables (15) and statements in my script (50) due to the things I shuffled in the last commit. Everything tested correctly now and this should be the last commit as long as you are happy with the changes I made!
Thanks :)

circuit_maintenance_parser/parsers/windstream.py
```diff
@@ -0,0 +1,78 @@
...
+            if header_tag is None or header_tag == "Maintenance Address:":
+                continue
+            header_tag = header_tag.string.strip()
+            value_tag = cols[1].string.strip()
+            if header_tag == "WMT:":
+                data["maintenance_id"] = value_tag
+            elif "Date & Time:" in header_tag:
+                value_tag = value_tag.replace(" ET", "")
+                date_obj_est = datetime.strptime(value_tag, "%m/%d/%y %H:%M")
+                date_obj_est = pytz.timezone("US/Eastern").localize(date_obj_est)
```
chadell, line 59: I guess that this may change depending on the customer's timezone. what do you think?

circuit_maintenance_parser/parsers/windstream.py
```diff
@@ -0,0 +1,78 @@
...
+                    data["end"] = int(date_obj_est.astimezone(pytz.utc).replace(tzinfo=timezone.utc).timestamp())
+            elif header_tag == "Outage":
+                impact = Impact("OUTAGE")
+            else:
+                continue
+
+        table = soup.find("table", "circuitTable")
+        for row in table.find_all("tr"):
+            cols = row.find_all("td")
+            if len(cols) == 9:
```
chadell, line 72: Do you think that the table will have always 9 columns? could we make this a bit more "robust"?

circuit_maintenance_parser/parsers/windstream.py
aliex-13: Hmm I guess that is a good point.
I may only have maintenance email examples from Windstream for our circuits that belong in that region. I haven't got any other emails to base this information on.

Do you have any functions which use a timezone checker that I can replicate (based on the abbreviation used at the end of a datetime string)?
If not, I'll come up with my own :)

circuit_maintenance_parser/parsers/windstream.py
aliex-13: We already make it a bit more robust by checking that it is a 'circuitTable'. From all the emails I have with Windstream, they keep those headers consistent and there is always 9 columns.

Do you have any recommendations to make this more robust rather than checking the length of the fields that we expect to be there?

circuit_maintenance_parser/parsers/windstream.py
chadell: no @aliex-13 , we don't have, but I think it would be a useful helper function

aliex-13: Hey @chadell - I've pushed a new commit with a new helper function in the "utils.py" file. I'm not sure if this was the correct location but I could see similar functions there. Please let me know if I need to move it/fix it.

I've added all of the basic US related timezones to this function. 
We can easily add more as shown in the code.

Does this work for you? All of my tests pass.

chadell (approved):
LGTM

Thanks @aliex-13 for your awesome contribution!

PR #288, opened 2024-06-04
Fix various exceptions under Pydantic 1.x

glennmatthews:
Although v2.6.0 claims to be backwards-compatible with Pydantic 1.x, using it with `nautobot-circuit-maintenance` with Pydantic 1.x causes it to throw exceptions in several places. This PR fixes those.


circuit_maintenance_parser/provider.py
```diff
@@ -173,7 +173,7 @@ def get_default_include_filters(cls) -> Dict[str, List[str]]:
             return cls._include_filter.get_default()  # type: ignore
```
scetron, line 173: does this need ()?

glennmatthews: no, the `()` (instantiating the class) is necessary in the pydantic 1.x case but appears not to be in the pydantic 2.x case.

scetron (approved):

PR #301, opened 2025-01-02
Add Pccw parser

snitass:
This code implements parsers for PCCW (a telecommunications company) circuit maintenance email notifications. It consists of two main classes:

HtmlParserPccw: Parses the HTML content of maintenance emails to extract:
- Provider name (PCCW Global)
- Customer account information
- Maintenance window start and end times

SubjectParserPccw: Parses the email subject line to extract:
- Maintenance ID
- Status (defaults to COMPLETED)
- Summary information

The parsers use regular expressions and string manipulation to extract relevant information from the email content. The datetime values are handled in a specific format ("%d/%m/%Y %H:%M:%S") and converted to timestamps.


circuit_maintenance_parser/parsers/pccw.py
```diff
@@ -0,0 +1,87 @@
...
+        return start_time, end_time
+
+
+class SubjectParserPccw(EmailSubjectParser):
+    """Custom Parser for Email subject of Pccw circuit maintenance notifications.
+
+    This parser extracts maintenance ID, status and summary from the email subject line.
+    """
+
+    DEFAULT_STATUS: ClassVar[Status] = Status.COMPLETED
```
chadell, line 59: shouldn't this be CONFIRMED? COMPLETED means that the maintenance has been executed already

circuit_maintenance_parser/__init__.py
```diff
@@ -26,6 +26,7 @@
     Momentum,
     Netflix,
     PacketFabric,
+    Pccw,
```
chadell, line 29: you could use PCCW in upper case, like NTT, if it matches better the company name

circuit_maintenance_parser/parsers/pccw.py
```diff
@@ -0,0 +1,87 @@
...
+        """Parse Pccw circuit maintenance email.
+
+        Args:
+            soup: BeautifulSoup ResultSet containing the email HTML content
+
+        Returns:
+            List containing a dictionary with parsed maintenance data
+        """
+        data: Dict[str, Any] = {
+            "circuits": [],
```
chadell, line 27: where are you taking the circuits' information? (it is mandatory to create the maintenance object)

tests/unit/test_e2e.py
```diff
@@ -640,6 +641,16 @@
                 GENERIC_ICAL_RESULT_PATH,
             ],
         ),
+        # Pccw
```
chadell, line 644: you could add the other e2e tests in ICAL format

chadell:
In the CI, there is an error due to Python 3.8 support:
```
    def _extract_maintenance_window(self, soup: ResultSet) -> tuple[datetime, datetime]:
E   TypeError: 'type' object is not subscriptable
```
-> The syntax tuple[datetime, datetime] is subscriptable typing, which is supported only in Python 3.9+.
Because Python3.8 is EOL, I'm gonna drop support in develop for the next release.

circuit_maintenance_parser/parsers/pccw.py
snitass: The completion notification doesn't come with ICAL. Others such as planned outage, urgent maintenance, amendment and cancellation notifications come with ICAL. That is why I set status to be `COMPLETED`.

chadell: that's a great explanation that would be worthy to add to the code to make future contributors aware of it

circuit_maintenance_parser/parsers/pccw.py
snitass: The circuit in completion notification is lack of some information than circuit in other notifications. For example, in planned outage notification (ICAL), circuit is `SR=SR654321:ABC(XXX)-DEF(XXX) EP4321:AAAAAA-654321-` but in completion notification (HTML), it's `SR654321 / ABC(XXX)-DEF(XXX) EP4321 /`.

Also, the following mentions some providers omit circuit in completion notification. Hence, I decided to leave circuit to be blank in completion notification.

>circuits: list of circuits affected by the maintenance notification and their specific impact. Note that in a maintenance canceled or completed notification, some providers omit the circuit list, so this may be blank for maintenance notifications with a status of CANCELLED or COMPLETED.

circuit_maintenance_parser/__init__.py
snitass: Done, thanks

tests/unit/test_e2e.py
snitass: Done, thanks

circuit_maintenance_parser/parsers/pccw.py
chadell: could you add a comment in the code about this?

snitass: Definitely

chadell (approved):
LGTM, many thanks @snitass !

PR #302, opened 2025-01-03
Drop Python 3.8 support and add Python 3.12

chadell:


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.6.1 and should not be changed by hand.
```
glennmatthews, line 1: What necessitated regenerating/updating the poetry.lock with new dependency versions?

jvanderaa (approved):

poetry.lock
chadell: https://github.com/networktocode/circuit-maintenance-parser/pull/302/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R25

jvanderaa: Not a bad thing to move poetry versions forward. Any harm that I'm not aware of on this?

glennmatthews: Looks like this is moving the `poetry` version itself backwards, which seems wrong. Just updating the min Python version shouldn't require re-locking all other dependencies (`poetry lock --no-update` would suffice); I'm not opposed to it in principle but it might be better done as a separate PR.

chadell: ah, I see what you mean, I shouldn't move back

chadell: I have regenerated with the latest poetry. I agree that a separate PR could have been better, but I would keep it together to simplify it for now.

jvanderaa (approved):

PR #312, opened 2025-03-03
Move epoch parser outside of LLM due to inconsistencies

aliex-13:
Was testing out the LLM parser and realised that there were a lot of inconsistencies passing in the same email multiple times.

The LLM API has a lot of issues converting the date string into epoch time.

I've messed around with the API directly by asking it questions like "Please convert the following time <> into an epoch string" (I wanted to avoid using a full email as I thought there might have been some special characters in the middle of the strings).
It still couldn't do it consistently. I also did checks prior to the epoch being converted and can see that the LLM can extract the datetime str correctly every time... It just couldn't do the conversion.

I asked the same question directly to the ChatGPT application on my local PC and this worked flawlessly, so there must be some difference between the API and the app (And yes, I have made sure the model I was using was the same on both).

After some discussions with @chadell directly, I have pulled the epoch conversion from the llm_question and made sure we get a datetime string from the API (I've tested this part of the LLM and it can do it very consistently). I've then written a function to do the conversion from datetime string into an epoch time myself.

From my testing, this seems very consistent and accurate.

### DIFF
--- a/circuit_maintenance_parser/parser.py
+++ b/circuit_maintenance_parser/parser.py
@@ -417,7 +417,7 @@ class LLM(Parser):
             return epoch_time
 
         except Exception as error_received:
-            logger.debug("Error parsing dates: %s", error_received)
+            logger.error("Error parsing dates: %s", error_received)
             raise ParserError from error_received
 
     def _get_start(self, generated_json: dict):
--- a/circuit_maintenance_parser/parser.py
+++ b/circuit_maintenance_parser/parser.py
@@ -7,6 +7,7 @@ import datetime
 import quopri
 from typing import Dict, List
 from email.utils import parsedate_tz, mktime_tz
+from dateutil.parser import isoparse
 import hashlib
 
 import bs4  # type: ignore
@@ -297,15 +298,15 @@ class LLM(Parser):
     _data_types = PrivateAttr(["text/html", "html", "text/plain"])
 
     _llm_question = """Please, could you extract a JSON form without any other comment,
-    with the following JSON schema (timestamps in EPOCH and taking into account the GMT offset):
+    with the following JSON schema (start and end times are datetime objects and should be displayed in UTC):
     {
     "type": "object",
     "properties": {
         "start": {
-            "type": "int",
+            "type": "datetime",
         },
         "end": {
-            "type": "int",
+            "type": "datetime",
         },
         "account": {
             "type": "string",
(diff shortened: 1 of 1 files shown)


circuit_maintenance_parser/parser.py
```diff
@@ -408,6 +409,20 @@ def _get_circuit_ids(self, generated_json: dict, impact: Impact):
...
+    def _convert_str_datetime_to_epoch(self, time_str):
+        try:
+            # Using isoparse for flexible ISO8601 parsing
+            time_dt = isoparse(time_str).astimezone(datetime.timezone.utc)
+            epoch_time = int(time_dt.timestamp())
+            return epoch_time
+
+        except Exception as e:
+            # Log the error message for debugging
+            print(f"Error parsing dates: {e}")
```
chadell, line 421: could we use the  `logger`?

circuit_maintenance_parser/parser.py
```diff
@@ -408,6 +409,20 @@ def _get_circuit_ids(self, generated_json: dict, impact: Impact):
...
 
+    def _convert_str_datetime_to_epoch(self, time_str):
+        try:
+            # Using isoparse for flexible ISO8601 parsing
+            time_dt = isoparse(time_str).astimezone(datetime.timezone.utc)
+            epoch_time = int(time_dt.timestamp())
+            return epoch_time
+
+        except Exception as e:
+            # Log the error message for debugging
```
chadell, line 420:
```suggestion
```

circuit_maintenance_parser/parser.py
```diff
@@ -408,6 +409,20 @@ def _get_circuit_ids(self, generated_json: dict, impact: Impact):
...
+            print(f"Error parsing dates: {e}")
+            # Optionally, you might raise a custom exception or return None values
+            raise
```
chadell, line 423:
```suggestion
            raise ParserError from e
```

aliex-13:
Thanks @chadell . Adding those fixes now.

circuit_maintenance_parser/parser.py
```diff
@@ -408,6 +409,17 @@ def _get_circuit_ids(self, generated_json: dict, impact: Impact):
...
 
+    def _convert_str_datetime_to_epoch(self, time_str):
+        try:
+            # Using isoparse for flexible ISO8601 parsing
+            time_dt = isoparse(time_str).astimezone(datetime.timezone.utc)
+            epoch_time = int(time_dt.timestamp())
+            return epoch_time
+
+        except Exception as error_received:
+            logger.debug("Error parsing dates: %s", error_received)
```
chadell, line 420: I think it deserves an error, no?
```suggestion
            logger.error("Error parsing dates: %s", error_received)
```

chadell (dismissed):
LGTM

chadell (approved):
LGTM, thanks for your contribution!

PR #328, opened 2025-08-07
Add LLM tokens used to MetaData, fix PCCW parser and text hook improvement

mkekez-SIE:
Raising this MR to cover a couple of different issues.

1. Noticed that LLM token usage was only included as a log. It would be great to return that in the metadata so it can be used  outside of the CMP library.
2. Noticed that the PCCW parser had some checks that could be improved so made these changes.
3. Was getting an error very often:

``Processor CombinedProcessor from GenericProvider failed due to: 'utf-8' codec can't decode byte 0xe9 in position 72: invalid continuation byte``

**What I discovered:**
Byte ``0xe9`` is typically the Latin-1 / ISO-8859-1 representation of é.

Resolution:

Fix the ``get_text_hook`` function to do improve robustness before returning the data:

```shell
    def get_text_hook(raw: bytes) -> str:
        """Can be overwritten by subclasses."""
        try:
            # Decode quoted-printable if needed
            decoded_bytes = quopri.decodestring(raw)

            # Auto-detect and decode
            result = from_bytes(decoded_bytes).best()
            if result is not None:
                return str(result)
            return decoded_bytes.decode("latin-1", errors="replace")
        except (UnicodeDecodeError, ValueError, TypeError, AttributeError):
            # Final fallback if all above methods fail
            return raw.decode("utf-8", errors="replace")
```

### DIFF
--- a/changes/328.housekeeping
+++ b/changes/328.housekeeping
@@ -1,0 +1,3 @@
+Add LLM tokens used to MetaData
+Fix PCCW parser
+Text hook improvement


chadell:
Could you add a file like [this](https://github.com/networktocode/circuit-maintenance-parser/blob/develop/changes/320.housekeeping)? (I have to add a check in the CI process yet)

chadell (dismissed):
Awesome! (before merging, please add the change file)

mkekez-SIE:
Hi @chadell , Changes doc has been pushed.
Please let me know if there is a specific structure to that file. I've added each specific thing fixed to a new line instead of on a single line.

PR #330, opened 2025-09-02
Update AWS parser for HTML emails

beetanz:
After updating the maintenance emails to go from AWS Health managed notifications they utilize an HTML format instead of plaintext. This was causing the parsing to fail. Adding support for the HTML emails from AWS health and maintaining previous code for plaintext (SNS) emails.

### DIFF
--- a/circuit_maintenance_parser/parsers/aws.py
+++ b/circuit_maintenance_parser/parsers/aws.py
@@ -62,29 +62,62 @@ class TextParserAWS1(Text):
         """
         data = {"circuits": []}
         impact = Impact.OUTAGE
-        maintenace_id = ""
+        maintenance_id = ""
         status = Status.CONFIRMED
-        for line in text.splitlines():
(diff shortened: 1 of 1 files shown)


itdependsnetworks:
I think the expectations is to add the sample to the tests, here is a prior PR that updated the capabilities of an existing parser that should be a good guide: https://github.com/networktocode/circuit-maintenance-parser/pull/256/files

beetanz:
Alright, I've got the tests in place and refactored the html parser into its own parser.

jvanderaa (changes requested):
Just the one minor change that we shouldn't be making.

beetanz:
I think it should be passing the tests now?

beetanz:
Is there anything else that is needed here?

jvanderaa:
@scetron @pke11y or @glennmatthews can you put some eyes on this to validate the changes that I have made. I'm good.

PR #331, opened 2025-09-08
Fix bug with maintenance_id generation due to CircuitImpact

mkekez-SIE:
Noticed many errors like the below:

`- Processor CombinedProcessor from GenericProvider failed due to: sequence item 0: expected str instance, CircuitImpact found`

After debugging, could see circuits were happy and looked good, so kept digging.

Found this function which is being used to generate a maintenance_id based on start/end combination + circuits into a hashed result (if a maintenance_id couldn't be extracted from the email).

Instead of iterating on the list of circuits which includes circuitImpact and other things, we should just be doing this on the generated_json circuits object list.

Testing locally shows this working as expected.

### DIFF
--- a/circuit_maintenance_parser/parser.py
+++ b/circuit_maintenance_parser/parser.py
@@ -500,7 +500,7 @@ class LLM(Parser):
 
         return account
 
-    def _get_maintenance_id(self, generated_json: dict, start, end, circuits):
+    def _get_maintenance_id(self, generated_json: dict, start, end):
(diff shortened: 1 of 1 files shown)


jvanderaa:
I was able to get a deeper dive on this. I agree with what I'm seeing that we should be getting a maintenance_id from the json not the individual circuit information.

mkekez-SIE:
Thanks guys!

Issue #334, opened 2025-09-19
OpenAI parser can't assess the Subject alongside the Body of emails


mkekez-SIE:
### Environment

- circuit_maintenance_parser version: 2.8.0

### Description

It seems that in some cases, our vendors send notifications which only contain the maintenance ID in the subject of the email, rather than the body (as well).

It has come to my attention that due to the 'subject' being a different datapart from the body of the email (whether it is text/html, text/plain or something completely different), it is never assessed with the rest of the information in the OpenAI Parser (never sent to the LLM API).

There are options which I went over (Option 3 is the preferred option):

**1. We could have added EmailSubjectParser to the list of CombinedProcessors when an OpenAI key is found.**

We could have added it last in the loop - if no maintenance-id was found by any other CombinedProcessor parser prior to it, then try to assess the subject of the email and 'try' to extract a maintenance ID. I guess this part was the difficult part because we were really guessing at what could be in the subject and where it could be located within the subject, so we went with a regex pattern of some sort. It was not a strong solution as there a many different variations which can be sent and a regex parser probably wasn't the best solution.
Another issue was the fact that other code within the library actually avoids this situation entirely. If no maintenance-id is found in the OpenAIParser, it will fallback to hashing a new maintenance ID using fields such as start/end time and affected services. So maintenance ID should always be filled in this scenario by the time we reach the **EmailSubjectParser**. Its hard to write logic that can overwrite this field is something always exists in the field. 

**2. Write individual HTML parsers for vendors which often do this (MaintenanceID only in subject of email)**

This is how I can see things were solved for other vendors in the CMP library.
It will take more hours and maintenance over time to go with this solution of creating a new vendor, html templates and tests. Not preferred in my opinion as there is one other solution which may work out better.

**3. Add the subject of the email to the body of the email (data parts which correlate to that).**

This is by far the best solution to allowing the OpenAI parser to actually use an LLM to assess the subject of the email. Adding the subject to the end of each datapart which could potentially be assessed will allow the LLM to have that extra data on hand and actually be a part of assessing whether the body of the email vs the subject is better for particular fields (e.g. the maintenance ID).

### Proposed Functionality

Before iterating over each processor in _get_maintenances_, we need to iterate over the NotificationData and see if we can find **email-header-subject**. If found, we need to extract this value, and then iterate on anything that has a content-type of 'text/*' or 'html' and append it to them.

```shell
[...]

PR #335, opened 2025-09-19
Add subject to body of emails for OpenAI parser

mkekez-SIE:
This should close https://github.com/networktocode/circuit-maintenance-parser/issues/334

Raising as a potential solution but keen for a review and potentially some other ideas/fixes.


jvanderaa:
Do we have any example/test data that we can update on the PR?

mkekez-SIE:
> Do we have any example/test data that we can update on the PR?

Hi @jvanderaa - I've added some tests to add the subject to existing dataparts.
This should cover what is required on the testing side of things.

Thanks!

PR #375, opened 2026-02-11
Improvements to Google Parser

AndriusV4:
Currently Google parser statically sets status to CONFIRMED for all maintenances. But Google sends emails with multiple statuses (Scheduled, Completed, Canceled). To cover this:

- Adding subject parser
- Moving `status` parsing to subject parser (HTML status parsing is cumbersome because 'Scheduled' status maintenances have no status in the main div, only 'Completed' and 'Canceled' do (and the status is in a separate <span> element, so current HTML parsing logic was failing on those)
- Moving `summary` and `maintenance_id` parsing to subject parser for simplicity 

Also Google sometimes send maintenances without any End Date specified at all, so to cover this:
- Added a check if no End Time was found
- Creating a dummy End Time of Star Time + 1hour, because Start/End times cannot be equal in this library.

Also added multiple tests to cover both parsers and e2e functionality as well.


glennmatthews (commented):
Thanks for the PR, looks solid at a quick readthrough. Would you mind adding a changelog fragment?

AndriusV4:
Thanks, added changelog fragment

changes/375.added
glennmatthews, line 1: Sorry for not providing more specific guidance earlier - this filename should be `changes/375.added`. Thanks.

AndriusV4: Apologies, fixed.

circuit_maintenance_parser/parsers/google.py
```diff
@@ -37,9 +38,42 @@ def parse_html(self, soup):
                 cid = googleaddr + "-" + span.parent.next_sibling.string.strip()
                 data["circuits"].append(CircuitImpact(circuit_id=cid, impact=Impact.OUTAGE))
 
-        summary = list(soup.find("div").find("div").strings)[-1].strip()
-        match = re.search(r" - Reference (.*)$", summary)
-        data["summary"] = summary
-        data["maintenance_id"] = match[1]
+        # Google sometimes send notifications without End Time specificed
```
jvanderaa:
```suggestion
        # Google sometimes send notifications without End Time specified
```

jvanderaa (dismissed):
Looking good with just one spelling in a comment suggestion.

jvanderaa (approved):
Looks good.

Issue #372, opened 2026-02-09
Enabling OpenAIParser breaks iCalendar parsing


AndriusV4:
### Environment

- Python version: 3.12.8
- circuit_maintenance_parser version: 2.10.0

### Observed Behavior
Enabling OpenAIParser causes ICal parser to fail parsing iCalendar format maintenance emails.

### Steps to Reproduce

1. Enable OpenAIParser to be added to the CombinedProcessor (by setting `PARSER_OPENAI_API_KEY` env variable)
2. Init any type of provider that sends emails in iCalendar format
3. Load an iCalendar email (The email *must not* have the exact subject string in the 'text/calendar' part to trigger this, but I have tested 3 different providers and none of them send the exact subject string in the iCal part)
4. Try to generate Maintenance objects from it

Example on how the iCal email is parsed fine, but it breaks as soon as we set `PARSER_OPENAI_API_KEY`:

```
circuit-maintenance-parser --data-file arelion.eml --data-type email --provider-type arelion

WARNING:circuit_maintenance_parser.provider:Not matching any exclude filter expression for Planned Work PWIC277299 Notification from Arelion.
Circuit Maintenance Notification #0
{
  "account": "*",
  "circuits": [
    {
      "circuit_id": "*",
      "impact": "OUTAGE"
    }
  ],
  "end": 1772546400,
  "maintenance_id": "PWIC277299.1",
  "organizer": "mailto:support@arelion.com",
  "provider": "arelion.com",
  "sequence": 8,
  "stamp": 1770376260,
  "start": 1772510400,
  "status": "CONFIRMED",
  "summary": "PWIC277299.1: Planned work notification from Arelion",
  "uid": "PWIC277299.1@arelion.com"
}

export PARSER_OPENAI_API_KEY="*"

circuit-maintenance-parser --data-file arelion.eml --data-type email --provider-type arelion 

WARNING:circuit_maintenance_parser.provider:Not matching any exclude filter expression for Planned Work PWIC277299 Notification from Arelion.
ERROR:circuit_maintenance_parser.parsers.openai:Expecting value: line 1 column 1 (char 0)
Provider processing failed: Failed creating Maintenance notification for Arelion.
Details:
- Processor SimpleProcessor from Arelion failed due to: Content line could not be parsed into parts: 'Planned Work PWIC277299 Notification from Arelion': Planned Work PWIC277299 Notification from Arelion
[...]

PR #376, opened 2026-02-12
Refactor provider to use OpenAI parser as a fallback and enhance test…

jvanderaa:
…s for processor behavior

Fix for the process of the OpenAI fallback to being a fallback. Includes a modification to not change the ical data.

### DIFF
--- a/circuit_maintenance_parser/provider.py
+++ b/circuit_maintenance_parser/provider.py
@@ -130,12 +130,6 @@ class GenericProvider(BaseModel):
             logger.debug("Skipping notification %s due filtering policy for %s.", data, self.__class__.__name__)
             return []
 
-        if os.getenv("PARSER_OPENAI_API_KEY"):
-            self._processors.append(CombinedProcessor(data_parsers=[EmailDateParser, OpenAIParser]))
-
-            # Add subject to all html or text/* data_parts if not already present.
-            self.add_subject_to_text(data)
-
         for processor in self._processors:
             try:
                 return processor.process(data, self.get_extended_data())
@@ -149,6 +143,22 @@ class GenericProvider(BaseModel):
                 error_message += process_error_message % related_exc
                 related_exceptions.append(exc)
                 continue
+
+        # Use OpenAI parser as a last resort if all other processors failed.
+        if os.getenv("PARSER_OPENAI_API_KEY"):
+            self.add_subject_to_text(data)
+            openai_processor = CombinedProcessor(data_parsers=[EmailDateParser, OpenAIParser])
+            try:
+                return openai_processor.process(data, self.get_extended_data())
+            except ProcessorError as exc:
+                process_error_message = (
+                    f"- Processor {openai_processor.__class__.__name__} from {provider_name} failed due to: %s\n"
+                )
+                logger.debug(process_error_message, traceback.format_exc())
+
+                related_exc = rgetattr(exc, "__cause__")
+                error_message += process_error_message % related_exc
+                related_exceptions.append(exc)
 
         raise ProviderError(
             (f"Failed creating Maintenance notification for {provider_name}.\nDetails:\n{error_message}"),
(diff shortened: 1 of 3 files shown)


circuit_maintenance_parser/provider.py
```diff
@@ -150,6 +144,22 @@ def get_maintenances(self, data: NotificationData) -> Iterable[Maintenance]:
                 related_exceptions.append(exc)
                 continue
 
+        # Use OpenAI parser as a last resort if all other processors failed.
+        if os.getenv("PARSER_OPENAI_API_KEY"):
+            self.add_subject_to_text(data)
```
glennmatthews, line 149: Still feels like modifying `data` in place may lead to bugs in the future. Might be better to change `add_subject_to_text(data)` so that it returns a modified copy instead of modifying the original in place?

tests/unit/test_providers.py
```diff
@@ -116,21 +116,78 @@ class ProviderWithIncludeFilter(GenericProvider):
...
+    with patch("circuit_maintenance_parser.processor.GenericProcessor.process") as mock_processor:
+        # All native processors fail, then OpenAI processor succeeds
+        mock_processor.side_effect = [ProcessorError] * original_processor_count + [[{"a": "b"}]]
+        provider.get_maintenances(data)
+        # Native processors + 1 OpenAI call
+        assert mock_processor.call_count == original_processor_count + 1
+
+    # OpenAI processor should NOT be appended to the provider's processor list
+    assert len(provider._processors) == original_processor_count  # pylint: disable=protected-access
+    os.environ.pop("PARSER_OPENAI_API_KEY", None)
```
glennmatthews: As a general pattern, it'd be nice to use [`patch.dict`](https://docs.python.org/3/library/unittest.mock.html#patch-dict) on `os.environ` as a way to guarantee it's cleaned up after each test, i.e:


```suggestion
    with patch.dict(os.environ, {"PARSER_OPENAI_API_KEY": "some_api_key"}):
        data = NotificationData.init_from_raw("text/plain", b"fake data")
        data.add_data_part("text/html", b"other data")

        provider = provider_class()
        original_processor_count = len(provider._processors)  # pylint: disable=protected-access

        with patch("circuit_maintenance_parser.processor.GenericProcessor.process") as mock_processor:
            # All native processors fail, then OpenAI processor succeeds
            mock_processor.side_effect = [ProcessorError] * original_processor_count + [[{"a": "b"}]]
            provider.get_maintenances(data)
            # Native processors + 1 OpenAI call
            assert mock_processor.call_count == original_processor_count + 1

        # OpenAI processor should NOT be appended to the provider's processor list
        assert len(provider._processors) == original_processor_count  # pylint: disable=protected-access
```

circuit_maintenance_parser/provider.py
jvanderaa: I think we should be good here, this is the last point in the path. Everything should be sent in and should be kept outside of the parsers.

jvanderaa:
Fixes #372

tests/unit/test_providers.py
```diff
@@ -116,21 +115,77 @@ class ProviderWithIncludeFilter(GenericProvider):
...
 
-    assert provider._processors[-1] == CombinedProcessor(  # pylint: disable=protected-access
-        data_parsers=[EmailDateParser, OpenAIParser]
-    )
+    os.environ.pop("PARSER_OPENAI_API_KEY", None)
+
+
+def test_provider_data_not_mutated_when_native_succeeds():
+    """Test that add_subject_to_text is not called when native processors succeed."""
+    os.environ["PARSER_OPENAI_API_KEY"] = "some_api_key"
```
glennmatthews: Can we change all of these test cases to use `patch.dict(os.environ, ...)` as was done for `test_provider_falls_back_to_openai` above?

glennmatthews (dismissed):

glennmatthews (approved):

PR #377, opened 2026-02-14
Fix Lumen parser status for scheduled maintenance

adam-hamilton:
Fixed a bug where scheduled future maintenance events were incorrectly being marked as IN-PROCESS instead of CONFIRMED.

The bug seems to have been introduced in commit 737aa4e9 (Aug 2021, PR #49 "Lumen Variations") when the parser logic was changed to set status to IN-PROCESS for emails containing either "The scheduled maintenance work has begun" OR "This maintenance is scheduled".

The fix separates these two conditions:
- "The scheduled maintenance work has begun" → IN-PROCESS (correct)
- "This maintenance is scheduled" → CONFIRMED (now correct)

Updated test expectations for lumen3, lumen4, and lumen8 which all contain "This maintenance is scheduled" text and represent future scheduled events.

### DIFF
--- a/changes/377.fixed.rst
+++ b/changes/377.fixed.rst
@@ -1,0 +1,1 @@
+Fixed Lumen parser incorrectly marking scheduled future maintenance events as IN-PROCESS instead of CONFIRMED. Bug introduced in commit 737aa4e9 (Aug 2021).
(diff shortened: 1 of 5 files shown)


glennmatthews (approved):
Seems reasonable to me!

jvanderaa (approved):

PR #378, opened 2026-02-23
Megaport fixes

i3dnet-akoopen:
Megaport fixes:

* Megaport sends initial maintenance announcements and reminder announcements. This parser was failing for the initial maintenance announcement which starts with "Please be advised that" instead of "This is a reminder". This fix should account for both emails.
* For their intial maintenance announcement the start and end date are not in separate paragraphs (in the reminder email this *is* the case). This fix should account for both emails.
* The "purpose of maintenance:"-section is used for the summary field of the maintenance. It seems that in newer emails the contents is being put in a separate paragraph. Account for both scenarios with this fix.

Lastly, I will also reach out to Megaport and ask if they can implement the BCOP standard as that seems more sensible then having to make these fixes :)


jvanderaa:
@i3dnet-akoopen can you take a look at the tests that are failing?

i3dnet-akoopen:
@jvanderaa hmm! could you give me some pointers on how I can resolve these tests failing?

itdependsnetworks:
You can run the tests locally, but they show up here: https://github.com/networktocode/circuit-maintenance-parser/actions/runs/22308122203/job/64560316105. 

So far there is:

1. pylint failing
2. need to add a changelog (see changes folder for example)
3. the parser is not matching expected results (i think an extra space)

i3dnet-akoopen:
@itdependsnetworks thanks for you input, this the first time for me sending a PR for this project. I think I should fixed those issues now, could you trigger a test-run again?

circuit_maintenance_parser/parsers/megaport.py
```diff
@@ -30,29 +30,43 @@ def parse_tables(self, tables: ResultSet, data: Dict):
...
+                    p_summary = False
                 elif p_text.startswith("Start Date and Time:"):
-                    re_search = re.search("Start Date and Time: (.*) UTC", p_text)
+                    # Megaport uses different formats in their initial maintenance announcement email and reminder email. In their reminder email they split start and end date across paragraphs
+                    re_search = re.search("Start Date and Time: (.*) UTC End Date and Time: (.*) UTC", p_text)
                     if re_search:
-                        start = parser.parse(re_search.group(1))
-                        data["start"] = self.dt2ts(start)
+                        data["start"] = self.dt2ts(parser.parse(re_search.group(1)))
+                        data["end"] = self.dt2ts(parser.parse(re_search.group(2)))
```
itdependsnetworks, line 61: With this change of looking for 2nd element, we should have more defensive programming, such as ensuring there is a `.group(2)` when calling on this.

i3dnet-akoopen: Hmm! What would you suggest the best way to check this? The same call is being made in multiple places in this parser, so we would have this fix all occurences then. In this commit I have only simply shuffled some code around to reduce the amount of local variables lower (and making the pylint test happy essentially 😅 ).

i3dnet-akoopen: I've added stricter checking now for this.

jvanderaa (approved):

jvanderaa:
Looking good. Thank you @i3dnet-akoopen

PR #380, opened 2026-03-04
Removes pylint.

### LINKED ISSUE #379: Remove pylint from CI and simplify pytest job to local execution
## Summary

The CI pipeline has unnecessary complexity that should be cleaned up:

1. **Pylint overlaps with ruff** — ruff already covers linting (flake8, pydocstyle, bandit, isort), making pylint redundant.
2. **Pylint and pytest CI jobs use Docker unnecessarily** — they build a Docker image inside the GitHub Actions runner just to execute commands, while ruff/yamllint already prove the simpler local-execution pattern works.
3. **`$$` bug in pylint job** — line 107 of ci.yml has `py$${{ matrix.python-version }}` which produces the shell PID instead of the Python version in image tags.

## Proposed Changes

- Remove `pylint` dependency from `pyproject.toml` and all `[tool.pylint.*]` config sections
- Remove `pylint` invoke task from `tasks.py` and its call from the `tests` aggregate task
- Remove the `pylint` CI job from `.github/workflows/ci.yml`
- Simplify the `pytest` CI job from Docker-based to local execution (matching the ruff/yamllint pattern)
- Regenerate `poetry.lock`

## Files Affected

- `pyproject.toml`
- `tasks.py`
- `.github/workflows/ci.yml`
- `poetry.lock`

jvanderaa:
## Summary

  - Removed pylint as a dev dependency and all [tool.pylint.*] config from pyproject.toml — ruff already covers linting
  - Removed the pylint invoke task from tasks.py and its call from the tests aggregate
  - Removed the pylint CI job from ci.yml, which also eliminates the $$ bug that produced the shell PID instead of the Python version in image tags
  - Simplified the pytest CI job from Docker-based execution to local execution, matching the pattern already used by ruff and yamllint
  - Regenerated poetry.lock

  Closes #379

## Test plan

  - `poetry run invoke ruff --action lint` # no new lint issues
  - `poetry run invoke ruff --action format` # all files formatted
  - `poetry run invoke yamllint` # CI YAML is valid
  - `poetry run invoke pytest` # 369 passed (6 pre-existing ATT xlsx failures, unrelated)
  - `poetry check --lock` # lock file consistent
  - CI passes on all three Python versions (3.10, 3.11, 3.12)


glennmatthews:
Feels like we should consider instead adopting `cookiecutter-ntc` standards for this repository rather than making one-off adjustments to its CI - thoughts?

glennmatthews:
As I commented on #379, I'm not sure that removing pylint altogether is desirable.

jvanderaa:
I'm onboard with adopting the `cookiecutter-ntc` standards. With where this repo was built it should have originated from an earlier version of them. So it should be easy. But I don't know that we have the easy method to adopt them. Let me see what I can do.

itdependsnetworks:
@jvanderaa, our good friend @jeffkala has built the tooling, should be pretty straight forward .

PR #394, opened 2026-03-17
Lxml version 602

huacchob:


circuit_maintenance_parser/parser.py
```diff
@@ -208,7 +208,7 @@ def parse_html(
     def clean_line(line):
         """Clean up of undesired characters from Html."""
         try:
-            return line.text.strip()
+            return line.text.strip().replace("\r\n", "\n").replace("\r", "\n")
```
glennmatthews, line 211: No further test data changes needed due to this fix?

glennmatthews (approved):

circuit_maintenance_parser/parser.py
huacchob: I added that fix because the older version of lxml returned \r\n by default, and tests were failing because of that. So that addition fixes that by standardizing the newlines

huacchob: I ran tests on lxml version 4.9.3 and 6.0.2 and both pass all existing tests

cheenpo:
thank you for this  ❤️

PR #399, opened 2026-04-16
Bump version to 2.11.1-beta

jvanderaa:
## Summary

Post-release version bump to `2.11.1-beta` on develop, per NTC release workflow.


glennmatthews (commented):
I think `2.11.1b1` would be more correct but this *may* be fine as-is.

jvanderaa:
@glennmatthews can you merge this in with it as it is? I can't seem to do so.

glennmatthews (approved):

PR #406, opened 2026-04-24
Bump pytest minimum to 9.0.3 (CVE-2025-71176)

jvanderaa:
## Summary

- Raises the minimum `pytest` constraint from `^9.0.0` to `^9.0.3` in `pyproject.toml` so consumers of this library and its dev environment can't install a vulnerable pytest.
- Refreshes `poetry.lock` (pytest `9.0.2` → `9.0.3`).
- Adds `changes/396.security` towncrier fragment.

## Context

Addresses [CVE-2025-71176](https://nvd.nist.gov/vuln/detail/CVE-2025-71176) / [GHSA-6w46-j5rx-g56g](https://github.com/advisories/GHSA-6w46-j5rx-g56g) — insecure `/tmp/pytest-of-{user}` tmpdir handling on UNIX (local DoS / privilege escalation). CVSS 6.8 (Medium). Dev-dependency only.

Supersedes #396 (Renovate's lockfile-only bump, blocked on `changelog` check).

### DIFF
--- a/changes/396.security
+++ b/changes/396.security
@@ -1,0 +1,1 @@
+Raised the minimum pytest version to 9.0.3 to address CVE-2025-71176 (insecure /tmp/pytest-of-{user} tmpdir handling on UNIX).
--- a/poetry.lock
+++ b/poetry.lock
@@ -2137,14 +2137,14 @@ extra = ["pygments (>=2.19.1)"]
 
 [[package]]
 name = "pytest"
-version = "9.0.2"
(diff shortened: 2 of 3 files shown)


changes/396.security
glennmatthews, line 1: Since this only affects the dev-environment and not the actual usage of this library, I think our normal convention would be to call this `.housekeeping` rather than `.security`.

glennmatthews (approved):

PR #407, opened 2026-04-24
Bump lxml minimum to 6.1.0 (CVE-2026-41066)

jvanderaa:
## Summary

- Raises the minimum `lxml` constraint from `>=4.6.2,<7` to `>=6.1.0,<7` in `pyproject.toml` so consumers of this library can't install a vulnerable lxml.
- Refreshes `poetry.lock` (lxml `6.0.2` → `6.1.0`).
- Adds `changes/400.security` towncrier fragment.

## Context

Addresses [CVE-2026-41066](https://nvd.nist.gov/vuln/detail/CVE-2026-41066) / [GHSA-vfmq-68hx-4jfw](https://github.com/advisories/GHSA-vfmq-68hx-4jfw) — XXE via default entity resolution in `iterparse()` and `ETCompatXMLParser`. CVSS 7.5 (High). lxml 6.1.0 flips the default to `resolve_entities='internal'`. Runtime dependency.

Supersedes #400 (Renovate's lockfile-only bump, blocked on `changelog` check).


glennmatthews (dismissed):
Seems desirable, but we may need to do some validation to make sure it doesn't cause incompatibilities with other libraries that haven't yet made a similar update. I did a quick search through the `nautobot` GH org and didn't see any immediate issues in that regard (Nautobot itself has recently made this update, though we haven't yet shipped the corresponding patch releases containing that update, they should be expected on Monday)

PR #412, opened 2026-05-01
Add parser for provider/vendor RETN.

i3dnet-akoopen:
This is a HTML/Subject parser for provider/vendor RETN (its a well-known European/Asian provider of wavelength/IP products).

Tests seem to pass (4 tests included in the PR):

383 passed in 12.27s 

circuit_maintenance_parser/parsers/retn.py                    62      7    89%

All tests have passed!


jvanderaa:
Could you match the Colt parser's approach for date parsing? RETN sends dates in DD.MM.YYYY format (e.g. 07.05.2026), which dateutil.parser.parse() ambiguously parses as MM.DD.YYYY when both numbers are ≤ 12, silently producing wrong timestamps. Colt handles the same European format by passing dayfirst=True (see circuit_maintenance_parser/parsers/colt.py:54-55). Updating the two parser.parse(...) calls in circuit_maintenance_parser/parsers/retn.py:103,105 to do the same will fix retn1 (07.05.2026 is currently being decoded as 2026-07-05 instead of 2026-05-07); retn2 and retn3 happen to round-trip correctly today only because their day numbers (24, 26) are >12 and force dateutil into DD.MM mode. Expected JSON for retn1 will need to be regenerated after the fix.

i3dnet-akoopen:
Oh! Nicely spotted... that should be fixed now. The test-result for "retn1" had to be adjusted indeed. Tests have passed succesfully.

jvanderaa (approved):
Looks good. Thank you.

PR #414, opened 2026-05-06
Add Telxius parser

i3dnet-akoopen:
Add Telxius parser... their emails are a bit basic in that sense that not much information can be extracted about the impact and status, but at least the information about CID's and start/end-date are there.

circuit_maintenance_parser/parsers/telxius.py                 73      2    97%

All tests have passed!


jvanderaa:
One small consistency fix before merge: on circuit_maintenance_parser/parsers/telxius.py:140, the default branch sets data["status"] = "CONFIRMED" as a raw string, while the three branches above it use the Status(...) enum (Status("RE-SCHEDULED"), etc.). Could you change it to data["status"] = Status("CONFIRMED") to match? Pydantic happens to coerce it today, but the inconsistency is easy to misread and could trip up future refactors.

i3dnet-akoopen:
Fair enough... thanks for spotting that; I fixed that now. Also re-ran tests, seem to work fine.

jvanderaa (approved):
Looks good! Thank you @i3dnet-akoopen

jvanderaa:
I will also work on having a release out this week.

PR #416, opened 2026-05-06
Add parser for Vodafone

i3dnet-akoopen:
Adding a parser for Vodafone (international).

=========================================================================================================== 382 passed in 12.14s ============================================================================================================

circuit_maintenance_parser/parsers/vodafone.py                72      2    97%

All tests have passed!

### DIFF
--- a/circuit_maintenance_parser/parsers/vodafone.py
+++ b/circuit_maintenance_parser/parsers/vodafone.py
@@ -2,12 +2,13 @@
 
 import logging
 import re
-from typing import Dict, List
+from typing import Any, Dict, List
 
+from bs4 import BeautifulSoup
 from bs4.element import ResultSet  # type: ignore
 from dateutil import parser
 
-from circuit_maintenance_parser.parser import Html, Impact, Status
+from circuit_maintenance_parser.parser import EmailSubjectParser, Html, Impact, Status
 
 logger = logging.getLogger(__name__)
 
@@ -15,9 +16,9 @@ logger = logging.getLogger(__name__)
 class HtmlParserVodafone1(Html):
     """Notifications Parser for Vodafone notifications."""
 
-    def parse_html(self, soup: ResultSet) -> List[Dict]:
+    def parse_html(self, soup: BeautifulSoup) -> List[Dict]:
         """Execute parsing."""
-        data: Dict[str] = {"circuits": []}
+        data: Dict[str, Any] = {"circuits": []}
         self.parse_crq(soup, data)
         self.parse_tables(soup.find_all("table"), data)
         self.parse_bold(soup.find_all("b"), data)
@@ -46,11 +47,11 @@ class HtmlParserVodafone1(Html):
                     elif "Services Affected" in col_mapping and col == col_mapping["Services Affected"]:
                         cid = td_elem.text.strip()
                     elif "Service Impact" in col_mapping and col == col_mapping["Service Impact"]:
-                        # not sure if other impact types exist, can be expanded of need-be
+                        # not sure if other impact types exist, can be expanded of need-be. Default to DEGRADED.
                         if "loss of service" in td_elem.text.lower():
                             impact = Impact("OUTAGE")
                         else:
-                            impact = Impact("OUTAGE")
+                            impact = Impact("DEGRADED")
                     col += 1
 
                 # at the end of the table row, add circuits to list, if defined
@@ -66,10 +67,14 @@ class HtmlParserVodafone1(Html):
         """
         window = 0
         for bold in bolds:
+            text_lower = bold.text.lower()
+
             # find start/end date/time
-            if (
-                data["status"] == Status("RE-SCHEDULED") and "new scheduled start" in bold.text.lower()
-            ) or "scheduled start" in bold.text.lower():
+            # in case the window is re-schedulded, the original and new window are listed; ignore original window
+            if "original scheduled start" in text_lower:
+                continue
+
+            if "scheduled start" in text_lower:
                 window_next = bold.next_sibling
                 while window_next:
                     text = window_next.text.strip()
@@ -77,8 +82,9 @@ class HtmlParserVodafone1(Html):
                         window = text
                         break
                     window_next = window_next.next_sibling
+
             # find summary
-            elif "description" in bold.text.lower():
+            if "description" in text_lower:
                 description_next = bold.next_sibling
                 while description_next:
                     text = description_next.text.strip()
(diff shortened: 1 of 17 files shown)


circuit_maintenance_parser/parsers/vodafone.py
```diff
@@ -0,0 +1,116 @@
...
+                    if "Customer" in col_mapping and col == col_mapping["Customer"]:
+                        data["account"] = td_elem.text.strip()
+                    elif "Services Affected" in col_mapping and col == col_mapping["Services Affected"]:
+                        cid = td_elem.text.strip()
+                    elif "Service Impact" in col_mapping and col == col_mapping["Service Impact"]:
+                        # not sure if other impact types exist, can be expanded of need-be
+                        if "loss of service" in td_elem.text.lower():
+                            impact = Impact("OUTAGE")
+                        else:
+                            impact = Impact("OUTAGE")
```
jvanderaa: We should have something else. Should we go with `DEGRADED` or `NO-IMPACT` here?

circuit_maintenance_parser/parsers/vodafone.py
```diff
@@ -0,0 +1,116 @@
...
+        <b>New Scheduled Start/End Date &amp; Outage Window:</b><br>
+        06/04/2026 00:00 to 13/04/2026 00:00 UTC <br>
+        """
+        window = 0
+        for bold in bolds:
+            # find start/end date/time
+            if (
+                data["status"] == Status("RE-SCHEDULED") and "new scheduled start" in bold.text.lower()
+            ) or "scheduled start" in bold.text.lower():
+                window_next = bold.next_sibling
```
jvanderaa: Here, the `OR` clause will match both of `scheduled start` is in the text body. So maybe we are looking something like:

```
  text_lower = bold.text.lower()
  is_new = "new scheduled start" in text_lower
  is_scheduled = "scheduled start" in text_lower

  if data["status"] == Status("RE-SCHEDULED"):
      if not is_new:
          continue
  elif not is_scheduled or is_new:
      continue
  # ... assign window, then break out of the for loop .
```

circuit_maintenance_parser/parsers/vodafone.py
```diff
@@ -0,0 +1,116 @@
...
+    """Notifications Parser for Vodafone notifications."""
+
+    def parse_html(self, soup: ResultSet) -> List[Dict]:
+        """Execute parsing."""
+        data: Dict[str] = {"circuits": []}
+        self.parse_crq(soup, data)
+        self.parse_tables(soup.find_all("table"), data)
+        self.parse_bold(soup.find_all("b"), data)
+
+        return [data]
```
jvanderaa: Let's look at something like this:

```
  def parse_html(self, soup: ResultSet) -> List[Dict]:
      """Execute parsing."""
      data: Dict[str, Any] = {"circuits": [], "status": Status("CONFIRMED")}
      self.parse_crq(soup, data)
      self.parse_bold(soup.find_all("b"), data)
      return [data]
```

Then drop the else: data["status"] = Status("CONFIRMED") fallback in parse_crq (line 115-116)

circuit_maintenance_parser/parsers/vodafone.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def parse_crq(self, soup: ResultSet, data: Dict):
+        """Vodafone maintenance_id's are in the format of CRQ[0-9] with 12 digits.
+
+        Before mentioning the CRQ, the status of the maintenance can be derived, for example:
+
+        Please be advised that the Planned Works have been Completed: CRQ000001312927
+        """
+        text = soup.get_text(separator=" ")
+        match = re.search(r"\b(.*)[\s:]+(CRQ\d{12})\b", text)
```
jvanderaa: Not a blocker for me on this one, but if we want to move to a small refactor to have a subject parser and less greedy regex.

**Suggested small refactor (library-idiomatic).** Vodafone's subject lines already carry status and the CRQ:

```
Subject: 3rd Party Essential Works CRQ000001325565 Affecting ACME CORP
Subject: Rescheduled 3rd Party Planned Works CRQ000001319054 Affecting I3
Subject: Completed Vodafone E2E Service Change Management Planned Works
```

This is the same shape `cogent.py`, `colt.py`, `globalcloudxchange.py`, and `aquacomms.py` already handle with an `EmailSubjectParser` subclass plus the HTML parser, glued together via `CombinedProcessor`.

Sketch:

```python
class SubjectParserVodafone1(EmailSubjectParser):
    """Parse status and (when present) the CRQ from the subject line."""

    def parse_subject(self, subject: str) -> List[Dict]:
        data: Dict = {}
        subject_lower = subject.lower()

        if "completed" in subject_lower:
            data["status"] = Status("COMPLETED")
        elif "rescheduled" in subject_lower:
            data["status"] = Status("RE-SCHEDULED")
        elif "postponed" in subject_lower or "cancelled" in subject_lower:
            data["status"] = Status("CANCELLED")
        else:
            data["status"] = Status("CONFIRMED")

        crq_match = re.search(r"\bCRQ\d{12}\b", subject)
        if crq_match:
            data["maintenance_id"] = crq_match.group(0)

        return [data]
```

Then in `provider.py`:

```python
class Vodafone(GenericProvider):
    _default_organizer = "networkchangemanagement@vodafone.com"
    _processors = [
        CombinedProcessor([EmailDateParser, SubjectParserVodafone1, HtmlParserVodafone1]),
    ]
```

With that in place, `HtmlParserVodafone1.parse_crq` can shrink to a CRQ-only anchored search (no `(.*)`, no status derivation, no upstream-text false-positives):

```python
def parse_crq(self, soup: ResultSet, data: Dict):
    text = soup.get_text(separator=" ")
    match = re.search(r"\bCRQ\d{12}\b", text)
    if match:
        data.setdefault("maintenance_id", match.group(0))
```

This also resolves another ask for free: the subject parser always sets `status`, so `parse_bold` never reads a missing key.

circuit_maintenance_parser/parsers/vodafone.py
```diff
@@ -0,0 +1,116 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class HtmlParserVodafone1(Html):
+    """Notifications Parser for Vodafone notifications."""
+
+    def parse_html(self, soup: ResultSet) -> List[Dict]:
+        """Execute parsing."""
+        data: Dict[str] = {"circuits": []}
```
jvanderaa: **Invalid type annotation**: `data: Dict[str] = {"circuits": []}` is not a valid annotation (Dict needs two params). Use `Dict[str, Any]` or `dict`.

circuit_maintenance_parser/parsers/vodafone.py
```diff
@@ -0,0 +1,116 @@
...
+
+from circuit_maintenance_parser.parser import Html, Impact, Status
+
+logger = logging.getLogger(__name__)
+
+
+class HtmlParserVodafone1(Html):
+    """Notifications Parser for Vodafone notifications."""
+
+    def parse_html(self, soup: ResultSet) -> List[Dict]:
```
jvanderaa: type hint says `soup: ResultSet`, but it is a `BeautifulSoup` object. `ResultSet` is what `find_all` returns.

jvanderaa (commented):
Just a few updates on this one, but we should be good to get it in quick.

circuit_maintenance_parser/parsers/vodafone.py
i3dnet-akoopen: Added DEGRADED for now as the default. I've looked through many emails, but in our case "Loss of Service" was always the value. Perhaps other Vodafone customers could chime in on other statuses.

circuit_maintenance_parser/parsers/vodafone.py
i3dnet-akoopen: In the new commit, I've taken a more simple approach by ignoring the "original schedulded" window. Also because after the re-factor with the email subject parser data["status"] would not be populated anymore.

circuit_maintenance_parser/parsers/vodafone.py
i3dnet-akoopen: I agree an email subject parser makes sense here... I've tried to squeeze everything into a HTML parser, but the result is indeed a bit greedy. In the next commit I've included this.

circuit_maintenance_parser/parsers/vodafone.py
i3dnet-akoopen: Fixed

circuit_maintenance_parser/parsers/vodafone.py
i3dnet-akoopen: Fixed

circuit_maintenance_parser/parsers/vodafone.py
i3dnet-akoopen: Status would now be set in the email subject parser; so not needed anymore.

jvanderaa (approved):
Thank you @i3dnet-akoopen, looks good.

PR #417, opened 2026-05-11
Add Cirion parser (Lumen fork)

i3dnet-akoopen:
Hoi! Adding a parser for Cirion... Lumen has divested their South American business a while back. They are sending maintenance notices in somewhat the same format as Lumen, but there seem to be a bunch of differences. I've started off with the Lumen parser and stripped out and changed some things.

===
circuit_maintenance_parser/parsers/cirion.py                  70      4    94%

All tests have passed!


changes/417.added
```diff
@@ -0,0 +1 @@
+Add Cirion parser (Lumen fork).
```
jvanderaa, line 1:
```suggestion
Added Cirion parser (Lumen fork).
```

i3dnet-akoopen:
@jvanderaa FYI: in case you are working on a new release, you could potentially also include this parser, but no rush from my end at least.

jvanderaa (approved):

PR #419, opened 2026-05-15
Add parser for FLAG (fka Global Cloud Xchange)

i3dnet-akoopen:
Globalcloudexchange (Gcx) has rebranded to FLAG and with this their email format has changed slightly. I have started off with the Gcx parser and made the required changes to make it work.

Potentially the current Gcx parser could be deleted, but I'm unsure what the best way forward for this would be.

Good to note: FLAG can send multiple windows in the same email; however, it is not trivial to de-duplicate the maintenance data with only the HTML data. The maintenance_id and status only exist within the email subject parser, so the HTML parser doesn't know. For now, the HTML parser will only take the latest listed window. From what I have seen so far, FLAG only lists one window in their emails, but the format of course suggests otherwise...


circuit_maintenance_parser/parsers/flag.py
```diff
@@ -0,0 +1,112 @@
...
+                # this table is listing windows (note: for now, we will only use the last listed window)
+                elif table_type == "windows":
+                    data["start"] = self.dt2ts(datetime.strptime(cols[1].text.strip(), date_format))
+                    data["end"] = self.dt2ts(datetime.strptime(cols[2].text.strip(), date_format))
+
+    def parse_paragraphs(self, paragraphs: ResultSet, data: Dict):
+        """Parse paragraph elements to find account and summary."""
+        for p in paragraphs:
+            for pstring in p.strings:
+                # print(f"hoi: {pstring}")
```
jvanderaa: Let's remove this comment.
```suggestion
```

circuit_maintenance_parser/parsers/flag.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        data: Dict[str, Any] = {"circuits": []}
+        self.parse_tables(soup.find_all("table", attrs={"border-collapse": "collapse"}), data)
+        self.parse_paragraphs(soup.find_all("p"), data)
+
+        return [data]
+
+    def parse_tables(self, tables: ResultSet, data: Dict):
+        """Parse table elements to find maintenance windows (start/end) and circuit ID's."""
+        date_format = date_format = "%d-%b-%Y %H:%M"
```
jvanderaa:
```suggestion
        date_format = "%d-%b-%Y %H:%M"
```

circuit_maintenance_parser/parsers/flag.py
```diff
@@ -0,0 +1,112 @@
...
+            subject (str): subject of email
+            e.g. 'FLAG | PE2025102750538 | Planned Event | Rescheduled'.
+
+
+        Returns:
+            List[Dict]: Returns the data object with maintenance_id and status fields.
+        """
+        data = {}
+        search = re.search(
+            r"^FLAG \| ([A-Z0-9]+) \| ([\w\s]+) \| ([\w\s]+)$",
```
jvanderaa: Dead regex capture groups 

```
r"^FLAG \| ([A-Z0-9]+) \| ([\w\s]+) \| ([\w\s]+)$" 
```
captures groups 2 and 3 but never reads them. Drop to `(?:...)` unless we have a spot for them.

jvanderaa:
Let's keep the old parser for now. And then we can create an issue here to clean it up in a period of time. I can do that as we get this merged.

circuit_maintenance_parser/parsers/flag.py
i3dnet-akoopen: Whups, leftover debugging, removed.

circuit_maintenance_parser/parsers/flag.py
i3dnet-akoopen: Oh I see that came over from the Globalcloudxchange parser, fixed.

circuit_maintenance_parser/parsers/flag.py
i3dnet-akoopen: Fixed

i3dnet-akoopen:
Included suggestions.

jvanderaa (approved):
LGTM

PR #428, opened 2026-08-06
Add Hawaiki Provider, Parser and tests

mkekez-SIE:
Hawaiki have recently updated their maintenance emailing processes where they no longer include all correspondences into their newest email response. At this point, we can now provide a HTML parser which can hopefully parse these emails. Right now we are fully depending on OpenAI parser for their emails and there are a lot of emails/updates which Hawaiki send which should classify as 'No Maintenances Found'.

I have written the provider, parser and tests and have also provided an include filter for Hawaiki to hopefully ignore the updates they are sending that do not actually have parseable maintenance details.

Note: Hawaiki send a contingency window in each maintenance they send, but to me it seems they use it very rarely. I have not parsed this information in my change, but prefer that we see how clients interact with Hawaiki with this parser, and if they see this contingency window needs to be parsed, we can make further improvements there.


circuit_maintenance_parser/__init__.py
```diff
@@ -66,6 +67,7 @@
     Google,
     GTT,
     HGC,
+    Hawaiki,
```
jvanderaa: Let's move this before HGC to keep alphabetical.

circuit_maintenance_parser/parsers/hawaiki.py
```diff
@@ -0,0 +1,73 @@
...
+            Service ID: HW-CID-12345
+
+            Maintenance Window
+            Start: 2026-August-05 20:00 UTC
+            End: 2026-August-05 21:00 UTC
+        """
+        data: Dict = {
+            "circuits": [],
+            "status": Status.CONFIRMED,
+            "account": "Unknown",
```
jvanderaa:
```suggestion
            "account": "Customer info unavailable",
```

Matching what is in apple.py:52 and bso.py:52

circuit_maintenance_parser/parsers/hawaiki.py
```diff
@@ -0,0 +1,73 @@
...
+        lines = text.splitlines()
+        in_maintenance_window = False
+
+        for line in lines:
+            line = line.strip()
+
+            match = re.match(r"Service ID:\s*(.+)", line)
+            if match:
+                circuit_id = match.group(1).strip()
+                data["circuits"].append(CircuitImpact(circuit_id=circuit_id, impact=Impact.NO_IMPACT))
```
jvanderaa: All circuits get Impact.NO_IMPACT unconditionally here, regardless of what the notification says. The email format already carries a Service impact: line that the parser walks past — so a genuinely service-affecting Hawaiki maintenance gets recorded as harmless. Downstream consumers (e.g. Nautobot circuit-maintenance) key off impact to decide whether to act, so this is worse than the current OpenAI-parser fallback for those emails.

The wrinkle is ordering: Service ID: appears above Service impact: in the body, so the single-pass loop can't know the impact at the moment it appends the circuit. Simplest fix is to resolve impact up front, since all circuits in a Hawaiki notification share one impact line;  same approach equinix.py:82 takes ("all circuits in the notification share the same impact").

I'd also default to OUTAGE rather than NO_IMPACT for unrecognized phrasing. OUTAGE is the library's own default in CircuitImpact, and it's the safe direction to be wrong in an over-cautious maintenance window costs someone an unnecessary look, an under-cautious one costs them an unexpected outage.

jvanderaa:
Here may help:

```
IMPACT_LINE_RE = re.compile(r"^Service impact:\s*(.+)$", re.IGNORECASE | re.MULTILINE)


    @staticmethod
    def _parse_impact(text: str) -> Impact:
        """Map the notification's single 'Service impact' line to an Impact."""
        match = IMPACT_LINE_RE.search(text)
        if not match:
            logger.warning("No Hawaiki 'Service impact' line found, defaulting to OUTAGE.")
            return Impact.OUTAGE
        impact_text = match.group(1).strip().lower()
        if "no service impact" in impact_text:
            return Impact.NO_IMPACT
        if "redundancy" in impact_text or "re-routed" in impact_text:
            return Impact.REDUCED_REDUNDANCY
        logger.warning("Unrecognized Hawaiki service impact %r, defaulting to OUTAGE.", impact_text)
        return Impact.OUTAGE
```

and in parse_text, replacing line 55:

```
        # All circuits in a Hawaiki notification share one "Service impact" line, and it appears
        # after the "Service ID" lines, so resolve it up front.
        impact = self._parse_impact(text)

        ...
            match = re.match(r"Service ID:\s*(.+)", line)
            if match:
                data["circuits"].append(CircuitImpact(circuit_id=match.group(1).strip(), impact=impact))
                continue
```

circuit_maintenance_parser/parsers/hawaiki.py
mkekez-SIE: Ahhh thats my bad! Will get that resolved.

mkekez-SIE:
@jvanderaa I think all threads should be resolved now! Please let me know if I am missing something.

jvanderaa (approved):
LGTM, thank you!

PR #433, opened 2026-08-17
Add KPN provider, xlsx parser and tests

### LINKED ISSUE #432: Add KPN as a Provider to the CMP library
### Environment
* circuit-maintenance-parser version:  2.12.0

### Proposed Functionality
Ability to parse KPN maintenance emails and attached CSVs.

### Use Case

KPN are a provider from the Netherlands which send all circuit information in CSVs. The AI parser has no chance to parse this information from the emails. The raw body of the email doesn't contain any useful information so created this parser. There is also some translation that was done as there is some dutch words in the emails/csv.

mkekez-SIE:
Resolves https://github.com/networktocode/circuit-maintenance-parser/issues/432

Added the relevant changes in order to introduce KPN as a provider to the CMP library.

All tests pass.

Only have a XLSX parser available as this is their method to notifications. Email body doesn't contain any useful information.

They do attach a PDF, but PDF parsers are not super supported yet.
Some Dutch in their emails to be aware of.


jvanderaa:
Thanks for this one @mkekez-SIE. CI is green across 3.10–3.14, the changelog fragment is in place, and the branch is current with `develop`. A few things I want to work through before I approve.

The one I care most about is the timezone. `_parse_datetime` stamps `tzinfo=timezone.utc`, but the spreadsheet has no timezone column and the email body doesn't state one. KPN is Dutch, so my read is those times are local (CET/CEST) and every window would land 1–2 hours off — at least from what I'm seeing in the fixtures. Can you confirm what timezone KPN publishes in? If it's local, `summitig.py` and `cogent.py` have the local→UTC pattern to follow.

Two more on the parser:

- `maintenance_id`, `status`, `start`, and `end` all come from `records[0]`. Does KPN ever put two `Astrid/IDA` values in one attachment? If so, we'd collapse two windows into one maintenance.
- The fixtures are synthetic (`Dummy Corp A`, `A000001`). One real notification, anonymized with `invoke anonymize-ips`, would let the tests prove the actual format — and would settle the timezone question at the same time.

Two small ones while you're in there:

- The title and description say "csv parser", but the code is `XlsxParserKPN1`. Worth correcting so this is findable later.
- Add KPN to the "Supported providers based on other parsers" list in `README.md`. (Hawaiki is missing from that list too, from #428 — I'll pick that one up separately, not yours to fix.)

The structure itself is in good shape: the Dutch status map, impact from downtime minutes, and the required-column check are all the right approach. Let me know if I've read the timezone piece wrong.

mkekez-SIE:
Hey @jvanderaa  - Thanks for taking a look at this.

Ahh yes - the timezone question you have is valid and I actually am not aware which timezone they are sending notifications in. In all of our notifications previously received, it is never stated which timezone they are sending these notifications in. I will follow up with their support team and get that answer.
Thanks for giving the examples of cogent and summitig!

> Does KPN ever put two Astrid/IDA values in one attachment? If so, we'd collapse two windows into one maintenance.

- I am not too sure as we have only ever received a single IDA in the attachment. I could update the code to capture this. Although, I might include this in my question to them when I reach out about the timezones!

> The fixtures are synthetic (Dummy Corp A, A000001). One real notification, anonymized with invoke anonymize-ips, would let the tests prove the actual format — and would settle the timezone question at the same time.

- The synthetic fixtures were actually generated using real examples of the emails we receive. There is no additional information in the emails to say which timezone they are referring too so I am not sure we could settle that without reaching out to them!

I'll also fix up those two little things! :)

Thanks!

mkekez-SIE:
Hi @jvanderaa - please take another look when you get the chance!

I spoke with KPN and they responded with the following:
 
- Does KPN ever put two Astrid/IDA values in one xlsx attachment?
_No, that is not possible, the IDA number is a unique maintenance number, every kind of maintenance has his own number._

- In which time zone is KPN sending these notifications? (CET/CEST or UTC)?
_We send the announcements in the CET/CEST time zone._

I copied parts of summitig and this should now be resolved so everything should be covered.
I added KPN and also Hawaiki to the README, so you don't have to worry about doing that!

Let me know if theres anything else missing!
Thanks

jvanderaa (approved):
Looks good with the updates. Merging

PR #434, opened 2026-08-19
Point documentation at docs.networktocode.com and fix site defects

### LINKED ISSUE #425: Read The Docs returning 404 error for project documentation
### Environment
* circuit-maintenance-parser version:  

### Expected Behavior
Attempted to view project documentation at https://circuit-maintenance-parser.readthedocs.io/

### Observed Behavior
Read The Docs returned a 404 'Project not found' error.

### Steps to Reproduce
1. View README.md
2. Click on any link in the 'Documentation' section.
3. Observe error.

jvanderaa:
Fixes #425

The Read the Docs Community project was retired, so every documentation link in the README and package metadata returns a 404. This repoints them, and the mkdocs `site_url`, at the Read the Docs for Business site under `docs.networktocode.com`.

A once-over of the site turned up several other defects, most of them cookiecutter leftovers that used the hyphenated project slug where the Python package name was required:

- **Broken header logo.** The theme pointed at `docs/assets/networktocode_logo.svg` while the file lived in `docs/images/`, so the logo 404'd. Moved the file (theme chrome belongs in `assets/`; `images/` is for content images embedded in pages).
- **"Edit this page" 404'd on every page.** `edit_uri` was `edit/main/circuit-maintenance-parser/docs`; the default branch is `develop` and docs live at `docs/`.
- **The Code Reference was silently empty.** `generate_code_reference_pages.py` globbed a `circuit-maintenance-parser` directory that does not exist, so the API docs the README advertises generated zero pages. It now globs the real package and renders as a single page under Developer Guide, using the existing `mkdocs-gen-files` — no new dependency.
- **Removed the unused `site_dir`,** which wrote the built site into a Nautobot-app `static/` path this library does not have.
- Fixed a v2.10 release-note link that used underscores in the repo name.

mkdocstrings `heading_level` moves 1 → 2 so each module becomes a heading below the page title, which restores the in-page table of contents.

Enabling the code reference surfaced seven griffe docstring warnings that fail the strict docs build; these are fixed at the source. `poetry.lock` is unchanged — this PR adds no dependencies.

## Note: this does not by itself make the docs reachable

The Read the Docs for Business subproject still needs to be marked **public** by an admin. Until that happens, `docs.networktocode.com/projects/circuit-maintenance-parser/` redirects to a login and the README docs badge reads "unknown".

## Upstream follow-up

Four of these defects come from the `python` template in `networktocode-llc/cookiecutter-ntc`, which uses `project_slug` where `project_python_name` (or a plain path) is required. Any repo generated from it likely has the same empty Code Reference and dead documentation links. Worth fixing upstream separately.


docs/generate_code_reference_pages.py
```diff
@@ -1,20 +1,15 @@
...
-    parts = list(module_path.parts)
-    if parts[-1] == "__init__":
-        parts = parts[:-1]
-
-    with mkdocs_gen_files.open(full_doc_path, "w") as fd:
-        identifier = ".".join(parts)
-        print(f"::: {identifier}", file=fd)
-
-    mkdocs_gen_files.set_edit_path(full_doc_path, file_path)
+with mkdocs_gen_files.open("code-reference.md", "w") as fd:
```
glennmatthews, line 9: So this changes from a `code-reference/` directory with one file per module into a single `code-reference.md` with all modules documented in a single file. Is that intentional/desirable?

pyproject.toml
```diff
@@ -4,9 +4,9 @@ version = "2.12.1b1"
 description = "Python library to parse Circuit Maintenance notifications and return a structured data back"
 authors = ["Network to Code, LLC <opensource@networktocode.com>"]
 license = "Apache-2.0"
-homepage = "https://circuit-maintenance-parser.readthedocs.io/"
+homepage = "https://docs.networktocode.com/projects/circuit-maintenance-parser/en/latest/"
```
glennmatthews, line 7: Do any of these fixes need to be additionally applied to the `.cookiecutter.json` file to make sure they don't get overwritten next time the package is rebaked?

jvanderaa: @jeffkala can you give me some insight here? From what I know and see we don't have this getting into cookiecutter today. I don't know the tool.

docs/generate_code_reference_pages.py
jvanderaa: Removed the change and filing it in another issue to take care of updating and standardizing the docs set up here.

PR #435, opened 2026-08-28
Add support for Telxius cancelled maintenance notices

i3dnet-akoopen:
Telxius sends cancelled maintenance notices in a different format (and different email subject)... great 😅 

Anyway, this patch should fix that. Passes the tests (new test added as well).

### DIFF
--- a/changes/435.fixed
+++ b/changes/435.fixed
@@ -1,0 +1,1 @@
+Add support for Telxius cancelled maintenance notices
--- a/circuit_maintenance_parser/parsers/telxius.py
+++ b/circuit_maintenance_parser/parsers/telxius.py
@@ -55,8 +55,16 @@ class HtmlParserTelxius1(Html):
             text = item.get_text(strip=True)
             # Remove optional notes like "(Backup Window)"
             text = text.split("(")[0].strip()
-            start_str, end_str = text.split(" - ")
-            events.append({"start": self.dt2ts(parser.parse(start_str)), "end": self.dt2ts(parser.parse(end_str))})
+            if "Start Time" in text:
+                start_str = text.split(": ")
+                start = self.dt2ts(parser.parse(start_str[1]))
+            elif "EndTime" in text and start:
+                end_str = text.split(": ")
+                end = self.dt2ts(parser.parse(end_str[1]))
+                events.append({"start": start, "end": end})
+            else:
+                start_str, end_str = text.split(" - ")
+                events.append({"start": self.dt2ts(parser.parse(start_str)), "end": self.dt2ts(parser.parse(end_str))})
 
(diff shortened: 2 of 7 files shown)


jvanderaa:
Thanks for this — and for chasing down yet another Telxius format variant. The fixture-plus-both-test-layers approach is exactly right.

One thing I'd like fixed before merge `start` can be read before it's bound

In parse_list_dates, start is referenced in the elif guard but only assigned inside the if branch. If an EndTime: bullet ever arrives without a preceding Start Time: bullet, that's an UnboundLocalError. It gets wrapped into ParserError upstream, so nothing crashes — but the operator sees a confusing traceback instead of "this schedule list is malformed". The related case worries me a bit more: because start is never reset after a pair is emitted, a Start / End / End sequence silently emits the same start twice rather than failing.

Worth flagging that neither linter catches this — I ran pylint with the repo's rcfile against your version of the file and it scored 10.00/10, so the green CI isn't evidence either way here.

While you're in there: "EndTime" is matched as a literal, missing space and all. That's Telxius's typo, and the day they fix it to End Time the bullet falls through to the legacy branch and dies on text.split(" - ") with ValueError: not enough values to unpack. Normalizing the label makes it survive that. Same for text.split(": "), which depends on the space after the colon.

Something like this covers all of it:

```python
def parse_list_dates(self, items: ResultSet, events: List):
    """Parse list elements to start and end datetime(s)."""
    start = None
    for item in items:
        text = item.get_text(strip=True)
        # Remove optional notes like "(Backup Window)"
        text = text.split("(")[0].strip()
        # Cancellation notices split the window into separate "Start Time:"/"EndTime:" bullets
        label, _, value = text.partition(":")
        label = label.replace(" ", "").lower()
        if label == "starttime":
            start = self.dt2ts(parser.parse(value.strip()))
        elif label == "endtime":
            if start is None:
                raise ValueError(f"Found an end time with no preceding start time: {text}")
            events.append({"start": start, "end": self.dt2ts(parser.parse(value.strip()))})
            start = None
        else:
            start_str, end_str = text.split(" - ")
            events.append({"start": self.dt2ts(parser.parse(start_str)), "end": self.dt2ts(parser.parse(end_str))})
```

I checked that partition(":") leaves the existing format alone — "12/August/2026 03:30 UTC - ..." produces a label of 12/august/202603, which falls through to else as before. Lowercasing the label also lines up with how parse_bold already matches ("notification number" in bold.text.lower()).

i3dnet-akoopen:
I've integrated the suggested changes into the parse_list_dates method.

jvanderaa:
Looks good. Thank you @i3dnet-akoopen

jvanderaa (approved):

